PR #6, opened 2016-08-11
Adding operation_description method to SwaggerRails::RSpec::DSL

singhgarima:
swagger ui supports "Implementation Notes" section, where we can write detailed notes for the corresponding API operation. Exposing `operation_description` method to allow integration specs to define content for "Implementation Notes" section. See image attached for outcomes.

screen shot 2016-08-12 at 8 26 32 am


Gemfile
```diff
@@ -19,3 +19,8 @@ group :development, :test do
   gem 'pry'
   gem 'generator_spec'
 end
+
+group :test do
+  gem 'test-unit'
```
domaindrivendev, line 24: why the need for test-unit here?

README.md
```diff
@@ -34,6 +34,7 @@ __NOTE__: It's early days so please be gentle when reporting issues :) As author
 
         post 'creates a new blog' do
           consumes 'application/json'
+          operation_description 'Detailed implementation notes for the create a new blog API endpoint'
```
domaindrivendev, line 37: I'd like to name this "description" rather than "operation_description" to be consistent with the other properties (consumes, produces etc.). The fact that they apply to an operation is implied by their position within the "post" context. My goal is for the terms and general structure of the DSL to align closely to those used by the swagger-spec itself, so that someone fluent in Swagger can pick up the DSL very quickly.

lib/swagger_rails/rspec/api_metadata.rb
```diff
@@ -0,0 +1,46 @@
+module SwaggerRails::RSpec
+  class APIMetadata
```
domaindrivendev, line 2: 👍 good call on the extraction

lib/swagger_rails/rspec/dsl.rb
```diff
@@ -20,6 +20,10 @@ def operation(http_verb, summary=nil, &block)
         describe(http_verb, metadata, &block)
       end
 
+      def operation_description(message)
```
domaindrivendev: rename to "description"

lib/swagger_rails/rspec/formatter.rb
```diff
@@ -17,12 +18,11 @@ def initialize(output)
       end
 
       def example_group_finished(notification)
-        metadata = notification.group.metadata
-        return unless metadata.has_key?(:response_code)
+        metadata = APIMetadata.new(notification.group.metadata)
+        return unless metadata.response_example?
```
domaindrivendev, line 22: 👏 cleans things up nicely

spec/dummy/spec/rails_helper.rb
```diff
@@ -29,7 +29,7 @@
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
   # instead of true.
-  config.use_transactional_fixtures = true
+  config.use_transactional_fixtures = false
```
domaindrivendev, line 32: I feel like the addition of DatabaseCleaner is unnecessary because there really shouldn't be any tests that actually call the API.

domaindrivendev:
First off - thanks! This is awesome with some great refactors. Just a few minor / nitpicky :) comments that would be great to address before I merge ...

Gemfile
singhgarima, line 24: @domaindrivendev when I was doing bundle exec rspec (without any changes), it was failing with error test-unit failing. 

```
bundler: failed to load command: rspec (/usr/local/var/rbenv/versions/2.3.0/bin/rspec)
LoadError: Please add test-unit gem to your Gemfile: `gem 'test-unit', '~> 3.0'` (cannot load such file -- test/unit/testcase)
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-3.2.22/lib/active_support/test_case.rb:2:in `require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-3.2.22/lib/active_support/test_case.rb:2:in `<top (required)>'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/railties-3.2.22/lib/rails/generators/test_case.rb:30:in `require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/railties-3.2.22/lib/rails/generators/test_case.rb:30:in `<module:Generators>'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/railties-3.2.22/lib/rails/generators/test_case.rb:9:in `<module:Rails>'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/railties-3.2.22/lib/rails/generators/test_case.rb:8:in `<top (required)>'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/generator_spec-0.9.3/lib/generator_spec/test_case.rb:2:in `require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/generator_spec-0.9.3/lib/generator_spec/test_case.rb:2:in `<top (required)>'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/generator_spec-0.9.3/lib/generator_spec/generator_example_group.rb:1:in `require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/generator_spec-0.9.3/lib/generator_spec/generator_example_group.rb:1:in `<top (required)>'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/generator_spec-0.9.3/lib/generator_spec.rb:2:in `require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/generator_spec-0.9.3/lib/generator_spec.rb:2:in `<top (required)>'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/runtime.rb:86:in `require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/runtime.rb:86:in `block (2 levels) in require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/runtime.rb:81:in `each'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/runtime.rb:81:in `block in require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/runtime.rb:70:in `each'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/runtime.rb:70:in `require'
  /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler.rb:102:in `require'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/dummy/config/application.rb:11:in `<top (required)>'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/dummy/config/environment.rb:2:in `require'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/dummy/config/environment.rb:2:in `<top (required)>'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/rails_helper.rb:3:in `require'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/rails_helper.rb:3:in `<top (required)>'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/swagger_helper.rb:1:in `require'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/swagger_helper.rb:1:in `<top (required)>'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/dummy/spec/integration/blogs_spec.rb:1:in `require'
  /Users/<path-in-my-local-to-project>/swagger_rails/spec/dummy/spec/integration/blogs_spec.rb:1:in `<top (required)>'
```

README.md
singhgarima, line 37: That makes sense. But description conflicts with ExampleGroup#description hence let me change it to implementation_notes which is how they show up in the swagger ui.

lib/swagger_rails/rspec/dsl.rb
singhgarima: renamed to "implementation_notes"

spec/dummy/spec/rails_helper.rb
singhgarima, line 32: I agree. My heart ached when I did this. But the new test I added for formatter didn't go well with the in built way of test data cleaning. It kept saying no active transactions to rollback. Which was kind of the right error. Hence I ended up switching to data cleaner which handled it better.

singhgarima:
@domaindrivendev did you got a chance to look at the new commit and comments I added?

Gemfile
domaindrivendev, line 24: I'd prefer to get to the bottom of this, rather than brute-forcing a dependency that really shouldn't be necessary. bundle rspec on master works fine for me.

We could add it for now just to get this PR accepted and consider it tech. debt to be addressed later?

singhgarima, line 24: @domaindrivendev I investigated a little more. It originates from gem generator_spec which depends on railties which depends on this gem.
Also I found:
"Ruby 2.2+ has removed test/unit from the core library. Rails requires this as a dependency."
And I am using ruby 2.3

README.md
domaindrivendev, line 37: Could we just go with "notes" on this one. I really want it to be "description" so that it reflects the terminology in the swagger-spec itself. Given that's not possible, I could settle on succinct alternative such as "notes".

domaindrivendev:
Hey there - thanks for investigating those few comments. If you could change "implementation_notes" to simply "notes" then I'd be happy to merge this in.

domaindrivendev:
@singhgarima what do you think about the following approach? A little hacky but would result in a more intuitive DSL

``````
```
# NOTE: This DSL should reflect the exact terms used in the swagger-spec
# See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md
# BUT: example group already has a 'description' method that takes no args
# As a workaround, override it and support both cases
def description(text=nil)  
  metadata[:operation_description] = text unless text.nil?
  super()
end
```
``````

domaindrivendev:
MERGED!

Hmmm - I just merged this in but the PR didn't automatically close. Must have missed a step somewhere but your commits are there. Thanks

drewish:
ah just found the PR and indeed it was working around the rspec naming.

drewish:
I'm noticing that this emits a `"description": null,` if you don't call `operation_description` in an operation. The [swagger editor](http://editor.swagger.io/) flags that as an error. Seems like it shouldn't be included if it's not called.

drewish:
Following up on my last comment https://github.com/domaindrivendev/swagger_rails/pull/18 should fix the null description issue.

PR #10, opened 2016-08-25
Keep resulting swagger.json consistent

dnswus:
Keep the execution order of the specs so that the resulting swagger.json will be consistent.


domaindrivendev:
makes sense - thx

PR #11, opened 2016-09-07
Quote the name in the Gemfile

drewish:
For the folks who cut and paste from the readme we should make the gem name a string, otherwise I think it'll be interpreted as a method name.


drewish:
@domaindrivendev this one seems like an easy merge. Any feedback?

PR #18, opened 2016-09-09
Try to ensure we generate valid swagger

drewish:
Swagger considers null values the `description`, `consumes`, `produces`, etc fields
as invalid. If I take what is currently in `spec/dummy/swagger/v1/swagger.json` and drop it in the [swagger editor](http://editor.swagger.io/) you'll get validation errors for the following:
- `description: null` => `Expected type array but found type null`
- `consumes: null` => `Expected type array but found type null`

These are easily fixed by rejecting the nil values.

### DIFF
--- a/spec/dummy/spec/integration/blogs_spec.rb
+++ b/spec/dummy/spec/integration/blogs_spec.rb
@@ -40,7 +40,7 @@ describe 'Blogs API', swagger_doc: 'v1/swagger.json' do
     get 'retrieves a specific blog' do
       operation_description 'For the id passed in the path, it searches and retrieves a blog against that id. If blog not found it returns a 404'
       produces 'application/json'
-      parameter :id, :in => :path, :type => :string
+      parameter :id, :in => :path, :type => :string, :required => true
 
       response '200', 'blog found' do
         let(:blog) { Blog.create(title: 'foo', content: 'bar') }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: ruby
+rvm:
+  - 2.2
+cache: bundler
+install: bundle install && cd spec/dummy && bundle exec rake db:setup && cd -
+script: bundle exec rspec
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 swagger_rails
 =========
 
+[![Build Status](https://travis-ci.org/domaindrivendev/swagger_rails.svg)](https://travis-ci.org/domaindrivendev/swagger_rails)
+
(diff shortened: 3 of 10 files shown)


drewish:
I went ahead and added the required property to the path param so it would validate as well.

drewish:
If you want another point of comparison here's their online validator that's less strict:
[sample file in master](http://online.swagger.io/validator/debug?url=https://raw.githubusercontent.com/domaindrivendev/swagger_rails/master/spec/dummy/swagger/v1/swagger.json)

``` json
{
  "messages": [
    "attribute paths.'\/blogs'(get).consumes is not of type `array`",
    "attribute paths.'\/blogs\/{id}'(get).consumes is not of type `array`"
  ],
  "schemaValidationMessages": [
    {
      "level": "error",
      "domain": "validation",
      "keyword": "allOf",
      "message": "instance failed to match all required schemas (matched only 0 out of 1)",
      "schema": {
        "loadingURI": "http:\/\/swagger.io\/v2\/schema.json#",
        "pointer": "\/definitions\/operation\/properties\/consumes"
      },
      "instance": {
        "pointer": "\/paths\/~1blogs\/get\/consumes"
      }
    },
    {
      "level": "error",
      "domain": "validation",
      "keyword": "type",
      "message": "instance type (null) does not match any allowed primitive type (allowed: [\"string\"])",
      "schema": {
        "loadingURI": "http:\/\/swagger.io\/v2\/schema.json#",
        "pointer": "\/definitions\/operation\/properties\/description"
      },
      "instance": {
        "pointer": "\/paths\/~1blogs\/get\/description"
      }
    }
  ]
}
```

[sample file in this branch](http://online.swagger.io/validator/debug?url=https://raw.githubusercontent.com/drewish/swagger_rails/7b0d8db711b2ee614da2716518345322be2bb23c/spec/dummy/swagger/v1/swagger.json)

``` json
{}
```

spec/dummy/spec/integration/blogs_spec.rb
```diff
@@ -40,7 +40,7 @@
     get 'retrieves a specific blog' do
       operation_description 'For the id passed in the path, it searches and retrieves a blog against that id. If blog not found it returns a 404'
       produces 'application/json'
-      parameter :id, :in => :path, :type => :string
+      parameter :id, :in => :path, :type => :string, :required => true
```
domaindrivendev, line 43: Given that path params are always required, would be convenient to have the generator set this automatically.

drewish: Agreed. I'll see what that change looks like next week

drewish: The generator will now add `required: true` for path parameters.

drewish:
I added tests for these changes.

drewish:
I rebased this on top of #22 so I could have Travis run and be sure the tests are all passing: https://travis-ci.org/drewish/swagger_rails/builds/159575519

PR #20, opened 2016-09-12
Capture request output as examples

drewish:
Curious what you think about this. I'd like to include the results of the test requests in the examples sections of the swagger docs. I made some kind of hacky changes to get the data back out but it works. I'll spend a little time getting it cleaned up into something more presentable and adding test coverage. 

One thing I'm not 100% sure how to handle is non-JSON responses. The spec only shows JSON examples which are shown being returned as objects. I'm not sure what you'd do with other formats... I suppose just return them as a string.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: ruby
+rvm:
+  - 2.2
+cache: bundler
+install: bundle install && cd spec/dummy && bundle exec rake db:setup && cd -
+script: bundle exec rspec
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 swagger_rails
 =========
 
+[![Build Status](https://travis-ci.org/domaindrivendev/swagger_rails.svg)](https://travis-ci.org/domaindrivendev/swagger_rails)
+
 Generate API documentation, including a slick discovery UI and playground, directly from your rspec integration specs. Use the provided DSL to describe and test API operations in your spec files. Then, you can easily generate corresponding swagger.json files and serve them up with an embedded version of [swagger-ui](https://github.com/swagger-api/swagger-ui). Best of all, it requires minimal coding and maintenance, allowing you to focus on building an awesome API!
 
 And that's not all ...
--- a/lib/swagger_rails/rspec/formatter.rb
+++ b/lib/swagger_rails/rspec/formatter.rb
@@ -6,6 +6,8 @@ module SwaggerRails
   module RSpec
     class Formatter
       ::RSpec::Core::Formatters.register self,
+        :example_finished,
+        :example_group_started,
         :example_group_finished,
         :stop
 
@@ -17,7 +19,32 @@ module SwaggerRails
         @output.puts 'Generating Swagger Docs ...'
       end
 
+      def example_group_started(notification)
+        @examples = nil
+      end
+
+      def example_finished(notification)
+        # TODO we should report errors so you know if you're trying to generate
+        # docs from failing specs
+        mime_type = notification.example.metadata[:response_mime_type]
+        body = notification.example.metadata[:response_body]
+
+        if mime_type && body
+          if mime_type == 'application/json'
+            begin
+              body = JSON.parse(body)
+            rescue JSON::ParserError => e
+            end
+          end
+
+          @examples = { mime_type => body }
+        end
+      end
+
       def example_group_finished(notification)
+        if notification.group.metadata[:response] && @examples
+          notification.group.metadata[:response][:examples] = @examples
+        end
         metadata = APIMetadata.new(notification.group.metadata)
         return unless metadata.response_example?
 
--- a/lib/swagger_rails/test_visitor.rb
+++ b/lib/swagger_rails/test_visitor.rb
@@ -17,6 +17,10 @@ module SwaggerRails
       else
         test.send(metadata[:http_verb], path, body_or_params, headers)
       end
+
+      # Store a copy of the results so we can use it as an example
+      metadata[:response_mime_type] = test.response.headers['Content-Type'].split(';').first
+      metadata[:response_body] = test.response.body
     end
 
     def assert_response!(test, metadata)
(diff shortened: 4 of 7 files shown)


drewish:
I re-ran the `rake swaggerize` task for the dummy rails site and here's the output:

``` json
{
  "swagger": "2.0",
  "info": {
    "title": "API V1",
    "version": "v1"
  },
  "paths": {
    "/blogs": {
      "post": {
        "tags": [
          "Blogs API"
        ],
        "summary": "creates a new blog",
        "description": "Creates a new blog. You can provide detailed description here which will show up in Implementation Notes.",
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "blog",
            "in": "body",
            "schema": {
              "type": "object",
              "properties": {
                "title": {
                  "type": "string"
                },
                "content": {
                  "type": "string"
                }
              }
            }
          }
        ],
        "responses": {
          "201": {
            "description": "valid request",
            "examples": {
              "application/json": {
                "errors": {
                  "content": [
                    "can't be blank"
                  ]
                }
              }
            }
          },
          "422": {
            "description": "invalid request",
            "examples": {
              "application/json": {
                "errors": {
                  "content": [
                    "can't be blank"
                  ]
                }
              }
            }
          }
        }
      },
      "get": {
        "tags": [
          "Blogs API"
        ],
        "summary": "searches existing blogs",
        "description": null,
        "consumes": null,
        "produces": [
          "application/json"
        ],
        "parameters": [

        ],
        "responses": {
          "200": {
            "description": "valid request",
            "examples": {
              "application/json": [

              ]
            }
          }
        }
      }
    },
    "/blogs/{id}": {
      "get": {
        "tags": [
          "Blogs API"
        ],
        "summary": "retrieves a specific blog",
        "description": "For the id passed in the path, it searches and retrieves a blog against that id. If blog not found it returns a 404",
        "consumes": null,
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "blog found",
            "examples": {
              "application/json": {
                "id": 1,
                "title": "foo",
                "content": "bar"
              }
            }
          },
          "404": {
            "description": "blog not found",
            "examples": {
              "application/json": "null"
            }
          }
        }
      }
    }
  }
}
```

It's actually pretty neat seeing that we're doing some odd things in that controller:
- returning an empty array for the index response — i guess some seed values would help there.
- returning a null value for 404s — it's not valid JSON and causes parse errors which was useful as a test case.

drewish:
I rebased this on top of #22 so I could have Travis run and be sure the tests are all passing: https://travis-ci.org/drewish/swagger_rails/builds/159576230

domaindrivendev:
This is interesting but IMO slightly contrary to the projects goal - namely that the "spec is the truth". You describe your API with metadata and it drives your integration tests and your documentation -"one spec to rule them all".

My plan for responses, which I just haven't implemented yet, is to add "schema" and "header" methods to the DSL so they can be used to describe payloads related to a given "response". Something along the following lines:

```
response '200', 'success' do
    schema type: 'object', ...
```

Then, I would have the run_test! method automatically assert the response against the provided schema.

Don't think this will be too hard to implement and hope to get to it soon

drewish:
I think the schema should definitely be in there there. I think the responses as examples are a separate feature.

domaindrivendev:
Fair enough, I'm just not overly keen on having implementation drive the spec.

domaindrivendev:
Btw ... thanks so much for all the PR's, I'm excited about this project but just haven't been able to give it the time I'd like. It's great to get all this help.

If you're interested we could sync up to agree on a feature roadmap, architecture, style guide etc. and then I could give you push access. Let me know?

drewish:
I appreciate the offer but I think I'm going to spend a little time doing a gem along similar lines but with a slightly different focus.

domaindrivendev:
"similar lines" ... I'll be interested to see how similar it is ;)

drewish:
oh i plan to copy the parts that i like ;)

stuff that's great:
- interface in the specs
- the rake driven workflow for generating static JSON files

things i'd want to try doing differently:
- being able to capture the output and incorporate that in the examples so they're more living docs is important for my use case
- i'll take the hackier option to overload `description` method so the DSL matches the swagger spec rather than changing names
- rake task should be running specs and reporting error during docs generation
- not worry about serving of swagger-ui and JSON
- much more granular testing and testing against against multiple versions of rails

odds are good it'll end up stalled under the weight of great ambitions, but i'll see what it looks like trying it with a clean slate.

domaindrivendev:
Well FWIW I would agree with all your goals bar injecting output into the spec. Re no need for UI, I plan to split JSON vs UI into two gems anyway.

Finally, if you are going to blatantly copy the general concept and I'm guessing code too, I would appreciate some form of reference. Cheers

drewish:
@domaindrivendev feel free to give https://github.com/drewish/rspec-swagger a look. it's still got a long ways to go in terms of polish but it's usable for my project. the API is very similar to your gem but the implementation is very different. it allows for a pretty full expression of the swagger spec:
- allows parameters at path item and object level
- uses consumes/produces elements at the document root
- an option to capture the result of a request as an example

one thing you might be interested in. i discovered you can pass both a body and query parameters by making the query params part of the path.

drewish:
One other thing you might be interested in is https://github.com/drewish/rspec-rails-swagger/pull/14 it's a generator to create a roughed out spec file for controller. Makes it much easier to add documentation to an existing code base.

domaindrivendev:
@drewish - you've created a competing gem which is a complete and utter rip off of this one. This in itself is very poor etiquette IMO (a fork would have been a more considerate approach), but that's OK - it's open source after all, people can do what they like. But seriously dude, could you at least refrain from pointing people to your gem from within this issue list?

drewish:
a fork would have made sense if it was just a few changes but there's literally none of the original your code in there. i can see how it would seem like poor etiquette to link to it but i was just trying to point a a few things that you might want to look at when you get some time to work on this. but sure i'm happy to stay out of your issue queue.

PR #22, opened 2016-09-12
Fix params

drewish:
Reverts the changes in https://github.com/domaindrivendev/swagger_rails/pull/9 and makes the change specific to Rails 5+. I'm not sure how you'd want to test the Rails 5 change... I don't see a great way to test that without adding a second rails instance or building some scripting to create the rails instance for testing.

I created this branch off of the Travis CI PR (https://github.com/domaindrivendev/swagger_rails/pull/21) to make sure it shows up as fixed. And based on https://travis-ci.org/drewish/swagger_rails/builds/159390811 it looks good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: ruby
+rvm:
+  - 2.2
+cache: bundler
+install: bundle install && cd spec/dummy && bundle exec rake db:setup && cd -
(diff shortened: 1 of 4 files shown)


drewish:
Closing this since I'll fix in the branch that adds Travis support so it can be tested across versions.

PR #32, opened 2016-10-18
Validating response headers and setting response examples

vinh0604:
This PR adds header validations into `Rswag::Specs::ResponseValidator`, using JSON::Validator to validate header value with [Header Object](http://swagger.io/specification/#headerObject)

I also add `response_examples` method to `Rswag::Specs::ExampleGroupHelpers` to support setting [response examples](http://swagger.io/specification/#exampleObject) into swagger docs.

Cheers,
Vinh Bach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ end
 
 ### Response examples ###
 
-You can provide custom response examples to the generated swagger file by calling the method `response_examples` inside the response block:
+You can provide custom response examples to the generated swagger file by calling the method `examples` inside the response block:
 
 ```ruby
 # spec/integration/blogs_spec.rb
@@ -275,7 +275,7 @@ describe 'Blogs API' do
     get 'Retrieves a blog' do
 
       response 200, 'blog found' do
-        response_examples 'application/json' => {
+        examples 'application/json' => {
             id: 1,
             title: 'Hello world!',
             content: '...'
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ In addition to paths, operations and responses, Swagger also supports global API
 # spec/swagger_helper.rb
 RSpec.configure do |config|
   config.swagger_root = Rails.root.to_s + '/swagger'
-  
+
   config.swagger_docs = {
     'v1/swagger.json' => {
       swagger: '2.0',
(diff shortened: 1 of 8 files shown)


domaindrivendev:
Hey there - awesome PR thx. I just have one ask though. Could you change the method name from "response_examples" to just "examples". I want the DSL to read exactly like Swagger. Just like "header" and "schema" your already in the context of a response so no need to qualify

vinh0604:
Hi @domaindrivendev, I actually tried with `examples` first, but then realized it overrides `examples` method of `Rspec::Core::ExampleGroup` [https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/example_group.rb#L433](https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/example_group.rb#L433) and causes `rspec` to fail:

```
./rswag/rswag-specs/lib/rswag/specs/example_group_helpers.rb:64:in `examples': wrong number of arguments (given 0, expected 1) (ArgumentError)
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/example_group.rb:481:in `next_runnable_index_for'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/metadata.rb:187:in `call'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/metadata.rb:187:in `build_scoped_id_for'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/metadata.rb:162:in `populate_location_attributes'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/metadata.rb:137:in `populate'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/metadata.rb:258:in `create'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/example_group.rb:420:in `set_it_up'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/example_group.rb:384:in `subclass'
    from ~/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.4/lib/rspec/core/example_group.rb:258:in `block in define_example_group_method'
```

domaindrivendev:
Gotcha - had the same issue with "description" and worked around by overriding the method to serve both use cases - one with an argument and one without. A little hacky but a trade-off I'm happy to make for a more consistent DSL. Checkout example_group_helpers#description

vinh0604:
Ah, I see. Haven't thought of that, I'll update accordingly. Thanks :D

vinh0604:
@domaindrivendev I pushed new commit for renaming `response_examples` to `examples` :)

On the second thought, I think we should avoid defining those DSL as global helper methods. As in this case, `examples` is actually a private class method of ExampleGroup, making our DSL implicitly depends on internal implementation of rspec.

We could change the DSL into something like:

``` ruby
response '200', 'OK' do |res|
  res.header ...
  res.examples ...
  res.schema ...
```

Or we may be able to keep the current usage as well as provide some level of isolation by using the similar technique for implementing DSL of `factory_girl` [https://robots.thoughtbot.com/writing-a-domain-specific-language-in-ruby](https://robots.thoughtbot.com/writing-a-domain-specific-language-in-ruby).

Any thought?

domaindrivendev:
I agree it's not ideal to be coupled to the internal implementation of rspec but everything in software is about trade offs and in this case, it's one I'm willing to make. I want the spec to read like a Swagger document and for this I feel the current DSL is just right.

Perhaps we reduce the risk by updating the Travis script to test against a few rspec versions including the most recent. But, all in all I'm happy to take the gamble and see what issues arise, if any

vinh0604:
That makes sense 👍 

In this case, anything else should I update for this PR?

domaindrivendev:
Something that's just occurred to me after re-reading the Swagger specification ... the [Swagger Header Object](http://swagger.io/specification/#headerObject) isn't really a JSON Schema. It's a subset of the fields and is specifically for describing data that would come back in a header string  not a JSON payload. So, using JSON Schema to validate these might present some issues. It will work fine when the header is described as a simple type (e.g. "string", "number", "integer", "boolean") but will fail for any headers described as type "array".

vinh0604:
@domaindrivendev I see, I also thought header values will be in JSON format when we set some array to it, but seems like it's not the case. I'll update it, maybe we need a custom HeaderValition class?

domaindrivendev:
Yep - a first cut could be just a private method in response_validator that just checks for the presence of required headers. Simple and maybe good enough.

If we want to get more fancy then a separate object to encapsulate validation rules for headers would make sense

PR #33, opened 2016-10-19
Simplify response header validation

vinh0604:
Change to checks for the presence of required headers instead of using JSON::Validator.

This is to fix the wrong assumption that header using JSON format to represent values introduced in https://github.com/domaindrivendev/rswag/pull/32


domaindrivendev:
Awesome thx!

PR #42, opened 2017-01-19
Add way for examples generation from responses

PavloBezpalov:
This changes add option to disable rspec 3 dry-run in generator. Without dry-run option its possible to extract response.body and write it as example to swagger files.


domaindrivendev:
awesome - simple and concise, plus I really appreciate the readme update too 👍

PR #46, opened 2017-02-16
add HTTP auth method to set basic auth

dan-ding:
When actually running these integration tests, it can be helpful to set the the http authentication header easily on various requests so that everything can complete. Or not set it (to test for a 401 response for instance). This is one way to do it.


Dahie:
Thank you, I'd also like to have this feature! 👍

MathieuGilbert:
I couldn't figure out how to get it to work with basic auth without these changes. Any plans to merge this in?

dan-ding:
@domaindrivendev is there something you'd like here so make it more appealing to merge?

domaindrivendev:
@dan-ding - so sorry for the delay in responding. Unfortunately the stuff that pays the bills gets takes precedence over the free open source projects :)

The [Swagger Spec.](http://swagger.io/specification/) includes a "securityDefinitions" field at the global level and a corresponding "security" field at the operation level to define security schemes (basic, api_key, oauth etc.) and then apply them to all or just specific operations. So, to remain consistent with the spec driven approach, I'd like to leverage this for specifying and applying basic auth requirements. This will also have the effect of enriching the generated documentation.

This is already in place for api key authentication, check out [test_app/spec/swagger_helper.rb](/domaindrivendev/rswag/tree/master/test-app/spec/swagger_helper.rb) and [test_app/spec/integration/blogs_spec.rb](/domaindrivendev/rswag/tree/master/test-app/spec/integration/blogs_spec.rb) to see this in action. I'd really like to maintain a consistent approach for basic auth too

I've just pulled down your branch to see what's involved in doing this. Stay tuned ...

And thanks for the support

domaindrivendev:
@dan-ding - take a look at #71 which takes the approach I had envisioned for this particular feature. Would this work for your needs?

dan-ding:
@domaindrivendev 
That, #71, appears to work -- though I do like the idea of a helper function of some sort:
basic_auth_or_some_helper(user, pass)
 instead of:
let(:Authorization) { "Basic #{::Base64.strict_encode64('jsmith:jspass')}" }

which ever way you want to take it.

domaindrivendev:
So here's my thoughts on the helper function ...

There's no question it provides some level of convenience. But I view specs as living documentation. They describe an API and provide example usages of that API. It could be a Ruby object or it could be a remote HTTP API as is the case rswag is addressing.

With that in mind, I would say the header format and base64 encoding is an important detail for the sample usage and therefore shouldn't be abstracted away behind some magic method.

dan-ding:
Ok! I feel ya! I really do. However I think well named function doesn't need hide anything except a lot of typing.
And the HTTP Basic Auth (and other auths) are liberally documented in throughout the world.
ActionController::HttpAuthentication::Basic.encode_credentials

domaindrivendev:
Fair points but ultimately we'll have to agree to disagree. I know I won't convince you (and likely others) because developers love their conveniences. Personally though I have zero issue with extra typing in favor of explicitness. Also if the redundancy really bothers application developers, they can just introduce their own helper.

One closing argument, in addition to the above, is that adding the helpers will pollute a certain consistency that I like about the current DSL. The way it works today - you describe you're API with the Swagger-style metadata then execute different examples by assigning values to specified parameters via rspec's "let" syntax. It's a simple and predictable approach that I'd like to maintain for all request parameters.

domaindrivendev:
Closing in favor of #71

MathieuGilbert:
Thanks @domaindrivendev this will help some people out!

dan-ding:
Fair enough!

Issue #44, opened 2017-02-13
Deprrecation warning from json-schema


shkrt:
`[DEPRECATION NOTICE] The preferred way to extend a Validator is by subclassing, rather than #extend_schema_definition. This method will be removed in version >= 3.`

https://github.com/ruby-json-schema/json-schema/blob/099357540ab40bb611b7e8d767c0d9f39032e4cc/lib/json-schema/schema/validator.rb#L16

domaindrivendev:
Fixed by #47

PR #47, opened 2017-02-16
get no deprecation; add some docs

dan-ding:
This updates the extended schema to use draft4 of json and not get the deprecation notice.
It also adds some documentation about the 'x-nullable' property that is available with the validator.


dan-ding:
This should fix #44

dan-ding:
@domaindrivendev if this isn't necessary anymore, would you close/comment or let me know if something else is needed?

domaindrivendev:
Awesome thx

Issue #53, opened 2017-04-12
how make upload request


thg303:
I tried this code to document an upload via put method, but it simply breaks. am I doing something wrong?

```path '/api/v1/puzzles/{puzzle_id}' do
     put 'update a puzzle details' do
      consumes: 'multipart/form-data'
      parameter name: 'Authorization', in: :header, type: :string, required: true
      parameter name: 'Content-Type', in: :header, type: :string, required: true
      parameter name: 'puzzle_id', in: :path, type: :string, required: true
      parameter name: 'puzzle[kind]', in: :formData, type: :string, enum: %w(freezed motioning), required: false
      parameter name: 'puzzle[picture]', in: :formData, type: :file, required: false
      response 200, 'successful' do
        let(:puzzle) { create :puzzle }
        let(:Authorization) { "Bearer #{get_token(puzzle.user)}"}
        let(:'Content-Type') { 'multipart/form-data' }
        let(:puzzle_id) { puzzle.id }
        let(:'puzzle[kind]') { 'freezed' }
        let(:'puzzle[picture]') { uploaded_file }
        run_test!
      end
    end
```

thg303:
mm.. looks like :formData is not implemented

raydog153:
Do you have a typo? for the parameter of type file you have in as 'formDate' and not 'formData'. I'm still trying to figure out how to set header and host for the request.

thg303:
@raydog153 oh yeah, that was a typo. but it does not work anyway.
for the header you may use my code, that works for simple queries.

domaindrivendev:
@thg303 - you're right form data not supported. I think it could be trivial enough - fancy making a PR?

pifleo:
Hello,

We had the same problem today and found out it was not implemented. We didn't manage to upload a file with `:body`.

So here is our patch to get a working file upload with `:formData` and `type: :file`:
```
# rswag-specs/lib/rswag/specs/request_factory.rb

module Rswag
  module Specs
    class RequestFactory

      # ...

      def build_body(example)
        if parameters_in(:formData).present?
          build_form_data(example)
        else
          body_parameter = parameters_in(:body).first
          body_parameter.nil? ? '' : example.send(body_parameter[:name]).to_json
        end
      end

      private

      def build_form_data_value(param, example)
        param[:type] == :file ? example.send(param[:name]) : example.send(param[:name]).to_s
      end

      def build_form_data(example)
[...]

PR #54, opened 2017-04-14
add formData support

thg303:
related to issue #53


pifleo:
Hey @thg303,

I had a look at your PR, and tested your code in place of [my patch](https://github.com/domaindrivendev/rswag/issues/53#issuecomment-294158739) in my code. 
It’s not working with a parameter of this form: `parameter name: :file, in: :formData, type: :file, required: true`.

From the Swagger spec it is also possible to have multiple `formData` parameters but only one` body` parameter is allowed, and your code in  ` request_factory.rb` doesn’t allow that as you use `first` on the param.
You put many parameter with formData in your spec but only the first ` 'blog[title]'
` is send with the request.

Your spec is interesting because the name of your params is a string with a scope, you should keep one like that.
You should keep a spec validating a parameter with the type `body`

Thanks for working on it

thg303:
thank you @pifleo , I refactored my PR to fix those issues

domaindrivendev:
@thg303 - thanks for this contribution. I'm still taking some time to go through it. One area that doesn't feel quite right, and may just be a limitation with Rails integration tests, is the way in which the client (i.e. the example) sends a Ruby hash for these form parameters. I would prefer it to simulate a "real" API client and send the body as a url-form-encoded string - similar to how the JSON body parameters are being submitted.

thg303:
@domaindrivendev I see what you mean, but unfortunately I could not come up with a real request simulation approach, as you mentioned seems to be a rails issue. by the way since the request is about to go to the rails, and it comes from rails, that would be an inner issue and could be ignored without any side effect.

citrusoft:
this is a very useful PR.  its been over two months.  why has this gone cold?  when are you planning to merge?

citrusoft:
Conlater.grats on the pull request acceptance.  Will be using this sooner
or

Tommy Hunt, CEA, PMP, PSM, SAFe Program Consultant
http://www.citrusoft.com
citrusoft@gmail.com
425-298-6551
Sent with Samsung Tab 3.

PR #57, opened 2017-04-28
Response body value validation

horiaradu:
Add the possibility to pass a block to the run_test!
method in order to add expectations on your response

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,6 +8,7 @@ PATH
   remote: ./rswag-specs
   specs:
     rswag-specs (1.2.0)
+      json (~> 1.8)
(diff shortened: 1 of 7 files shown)


domaindrivendev:
This looks awesome and I really appreciate the PR completeness - read me, specs etc.

Today, it is possible to do this by adding additional "it" blocks immediately after runtest! But, your solution is a little cleaner.

One minor change I might suggest would be to have the block accept the full response object instead of just the body. Just to give a little more flexibility. Thoughts?

domaindrivendev:
This is still just the body - only not parsed. I meant the entire response object so you could assert against the body, headers, status etc.

PR #59, opened 2017-05-18
Relax Rails dependency to allow 5.1

grk:


domaindrivendev:
Released as 1.3.0. Thanks for the contribution!

PR #61, opened 2017-05-26
Allow parameters to be defined only by schema references

djcp:
This should allow for larger schemas to be defined in a more compact fashion - please see the `swagger_helper` and integration spec changes in the test-app.

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -36,7 +36,9 @@ module Rswag
       end
 
       def parameter(attributes)
-        attributes[:required] = true if attributes[:in].to_sym == :path
+        if attributes.fetch(:in, nil) == :path
+          attributes[:required] = true
+        end
 
         if metadata.has_key?(:operation)
           metadata[:operation][:parameters] ||= []
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -43,12 +43,16 @@ module Rswag
 
       private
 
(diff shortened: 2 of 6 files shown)


djcp:
I also made sure the new swagger.json validates OK. It does!

djcp:
It'd be nice if there was a `CONTRIBUTING.md` that articulates how to run tests.

domaindrivendev:
Can't remember when/how this was added but looks like this functionality is already available. Please checkout 1.5.0 and let me know

djcp:
This does not appear to be working in 1.50 (sorry for the long delay on this).

```
...
path "/api/v3/addons/{uuid}" do
  let(:uuid) { addon.uuid }
  get "Get an add-on" do
    parameter(
      "$ref" => "#/parameters/uuid_path_param"
    )
...
```
yields the following error message:

```
2) V3 Addons API /api/v3/addons/{uuid} get Get an add-on by uuid. Herokai can see all add-ons. returns a 200 response
     Failure/Error: raise "Referenced parameter '#{ref}' must be defined" unless definitions && definitions[key]
     
     RuntimeError:
       Referenced parameter 'parameters/uuid_path_param' must be defined
```

This would be a useful feature and is supported by the swagger spec. Shall I rebase against 1.5.0 and re-issue or am I missing something?

domaindrivendev:
How are you defining your referenced parameter in swagger_helper.rb? If you use a string instead of a symbol as the key, it will likely work. If you could confirm this, it would be very helpful?

```ruby
parameters: {
  'uuid_path_param' => { ... }
}
```

Of course this is inconsistent with the other global metadata in swagger_helper and so I think it would be better to expect a symbol. So, I've updated the code in master. If you can confirm the above works, then I'll go ahead and publish the updated version so you can revert to the symbol approach.

PR #64, opened 2017-06-01
Add proper swagger ui url

jume-dev:
As this only supports swagger ui 2, the url should link accordingly


domaindrivendev:
Makes sense - thx

PR #65, opened 2017-06-06
Add Swaggerable to allow dynamic definition for model

moserke:
I don't know if something like this would be useful to the project.  I ended up writing this to close the last mile for our application where the schema could drift from the model.

Basically with this in the model I can do a `include Rswag::Models::Swaggerable`.  Then I can call `Model.swagger_definition` on the model to get the definition dynamically for that model.  This is used then in my swagger_helper.rb in the swagger_doc as 

```
definitions: {
  model: Model.swagger_definition
}
```

One main assumption made is that all models being swaggerized have a corresponding serializer.

There is also a helper method to be able to override found attributes with a different type. This was primarily used when using `serialize` as the underlying type is always a string and that's not what the actual type is. So in the model there is a helper method called `swagger_attribute(attribute, type)` that will override what the introspection finds.  This is also necessary when using an enum.

Let me know if this sounds interesting and if so what other things you'd want cleaned up before merging.  If it's not interesting, no worries, but it really helped make our tests more coupled to implementation so schema drift did not occur so thought I would share back.

### DIFF
--- a/rswag/lib/rswag.rb
+++ b/rswag/lib/rswag.rb
@@ -1,5 +1,6 @@
 require 'rswag/version'
 require 'rswag/specs'
+require 'rswag/models/swaggerable'
 require 'rswag/api'
 require 'rswag/ui'
 
--- a/rswag/lib/rswag/models/swaggerable.rb
+++ b/rswag/lib/rswag/models/swaggerable.rb
@@ -1,0 +1,76 @@
+module Rswag
+  module Models
+    module Swaggerable
+      extend ActiveSupport::Concern
+
+      included do
+        @swagger_custom_property_definitions = {}
+      end
+
+      class_methods do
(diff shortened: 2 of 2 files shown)


djcp:
We were just talking about how handy this'd be. :+1:

domaindrivendev:
Hey there thanks for submitting this but I'm going to decline. I don't think it's a bad idea - it's just not quite aligned with my vision for rswag ...

With rswag, the goal is for the specification to drive the implementation, not the other way around. That specification is at the HTTP/REST API level of abstraction, using a Swagger style DSL as the description language of choice. So, it's essentially TDD at the API integration layer.

This means you would define the behavior you expect from your API, without giving any thought to how it's implemented under the hood - maybe you use ActiveSupport::Validations, maybe you use something else. Then, you tweak the implementation to satisfy the specification.

PR #69, opened 2017-06-19
Fixed issue with precomilation

lazebny:
This PR resolves issue https://github.com/domaindrivendev/rswag/issues/68

I am not a big expert in swagger-ui, but what found in:

* https://github.com/d4be4st/swagger-ui_rails/blob/master/lib/swagger-ui_rails/engine.rb
* https://github.com/kendrikat/grape-swagger-ui

o2c.html is outside precompilation


domaindrivendev:
Great catch & thx

PR #75, opened 2017-07-06
adds request spec generator (shamelessly stollen from rspec-rails-swagger)

paulccarey:
Hi, 

I was flicking between rswag and rspec-rails-swagger, but decided to go with rswag because it looked more mature... however as I was looking at the two gems I noted the handy generator that @drewish had created for rspec-rails-swagger (Big Thanks!).

Anyways I thought it would be super handy to have in rswag... (I have a lot of end-points to document, so anything that helps me churn through them is great!)

Its very rough and ready right now (no tests) as I've almost literally just cut and paste the code from one gem to the other!

If this is a goer I'll tidy up the namespaces and add specs, let me know if you guys would like this adding as a permanent feature, I didn't want to invest time on tidying it up until someone says yea or nay.

To use its its simply a matter of 
```
rails generate rspec:swagger PostsController
``` 
example also shamelessly stollen from rspec-rails-swagger :trollface:


drewish:
Awesome, please take what ever code or inspiration you can from my code.

paulccarey:
Thanks @drewish 👍

domaindrivendev:
@paulccarey - haven't drilled too deep into the code but def in favor of adding this feature.

Also, I wouldn't worry about taking code from rspec-rails-swagger as the majority of that code base was shamelessly stolen from here :)

drewish:
The idea was shamelessly lifted. The code was not ;)

drewish:
Though if you're pulling stuff over I'd grab the tests and Travis config too. The routing code differs quite a bit between versions of Rails and what's there was designed to be compatible with versions 3-5.

rbetancor:
Seems to be a perfect addition to help noobs on RSpec, as me, I create the controllers and this create the specs to build the swagger doc.  :-)

rbetancor:
Found something weird on the generated rspec, check this simple controller:

```
class V1::DomainsController < V1::BaseController
  before_action :set_domain, only: [:show, :edit, :update, :destroy]

  # GET /domains
  def index
    @domains = Domain.all
  end

  # GET /domains/1
  def show
  end

  # GET /domains/new
  def new
    @domain = Domain.new
  end

  # GET /domains/1/edit
  def edit
  end

  # POST /domains
  def create
    @domain = Domain.new(domain_params)

    if @domain.save
      redirect_to @domain, notice: 'Domain was successfully created.'
    else
      render :new
    end
  end

  # PATCH/PUT /domains/1
  def update
    if @domain.update(domain_params)
      redirect_to @domain, notice: 'Domain was successfully updated.'
    else
      render :edit
    end
  end

  # DELETE /domains/1
  def destroy
    @domain.destroy
    redirect_to domains_url, notice: 'Domain was successfully destroyed.'
  end

  private
    # Use callbacks to share common setup or constraints between actions.
    def set_domain
      @domain = Domain.find(params[:id])
    end

    # Only allow a trusted parameter "white list" through.
    def domain_params
      params.require(:domain).permit(:name, :reseller_id)
    end
end

```

If I run the rails generate rspec:swagger V1::DomainsController the created rspec file fails with:

```
An error occurred while loading ./spec/integration/v1/domains_spec.rb.
Failure/Error: parameter 'id', in: :body, type: :string

ArgumentError:
  wrong number of arguments (given 2, expected 1)
# /usr/local/rvm/gems/ruby-2.3.3/gems/rswag-specs-1.3.0/lib/rswag/specs/example_group_helpers.rb:38:in `parameter'
# ./spec/integration/v1/domains_spec.rb:40:in `block (2 levels) in <top (required)>'
# /usr/local/rvm/gems/ruby-2.3.3/gems/rswag-specs-1.3.0/lib/rswag/specs/example_group_helpers.rb:7:in `path'
# ./spec/integration/v1/domains_spec.rb:38:in `block in <top (required)>'
# ./spec/integration/v1/domains_spec.rb:3:in `<top (required)>'
```

And taking a look at the generated rspec:

```
  path '/api/v1/domains/{id}/edit' do
    # You'll want to customize the parameter types...
      parameter 'id', in: :body, type: :string
      get('edit domain') do
      response(200, 'successful') do
              let(:id) { '123' }

        after do |example|
        end
        run_test!
      end
    end
  end

```
Why the parameter sentence said that "in: :body"? when its clear that the param goes "in: :path", either way it fails to 'swaggerize' the only way it pass is if I coment the parameter line.

Any clue on this? ... I'm very noob on RSpec and maybe i'm doing something really stupid here.

paulccarey:
@rbetancor its only a generator so it'll only really create you a skeleton to accommodate the most common use case which is a JSON style API with JSON payloads. 
You will still need to inspect the output and amend it as appropriate for your use case.

BookOfGreg:
I'm going to accept this but I'm going to tweak this a bit before release.
~The rspec namespace doesn't feel like something I want to use so I'll move this to `rails generate rswag:spec PostsController`~ EDIT: I was wrong, I prefer `rspec:swagger` after playing with it a while.
This should also have basic generator tests so I'll see about adding those in.
I'll also have a quick manual test of various versions of Rails to ensure how it uses the routes is good in all versions.

Thanks @paulccarey and @drewish for your contribution and patience!
Lets get this show on the road again 😄

PR #76, opened 2017-07-11
Support optional query parameters

ahobson:


rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -20,6 +20,7 @@ def build_fullpath(example)
 
       def build_query_string(example)
         query_string = parameters_in(:query)
+          .select { |p| p.fetch(:required, true) }
```
domaindrivendev: won't this mean that optional parameters will NEVER be added to the query string, regardless of a value being provided or not?

ahobson: Argh, yes.  Will fix in an updated PR.

Thank you

ahobson: PR updated.

PR #87, opened 2017-07-31
Allows for parameters to be defined without the 'in' key defined.

travishooper:
I'm new to using Swagger docs, however I have seen in some of my companies existing Swagger docs that they're using a $ref to define their parameters that get used a lot. I however cannot generate Swagger docs like this with this gem as it errors out when a `:in` parameter key isn't defined. This PR should fix that issue.


domaindrivendev:
Good catch - feature was 99% there but this change was crucial for it to actually work :)

Issue #88, opened 2017-08-03
Unable to use apiKey pairs in requests


daniel-sullivan:
As per: https://swagger.io/docs/specification/authentication/api-keys, It should be possible to use multiple apiKey entries together simultaneously. Ie: App Key + API Key.
However in testing we've discovered rswag's spec helpers only add the first entry of type apiKey to the headers / query.

Here is our securityDefinitions / security section showing an expected working use case:

      "securityDefinitions": {
          username: {
              type: :apiKey,
              name: "X-Username",
              in: :header
          },
          key: {
              type: :apiKey,
              name: "X-ApiKey",
              in: :header
          }
      },
      "security": [
          {
              "username": [],
              "key": []
          }
      ],

The swagger.json file is generated correctly. It's only affects the rspec tests.

I've added some tests and fixed the issue and will put in a pull request shortly.

domaindrivendev:
I believe this is already supported and it's just your syntax for the global security requirements is incorrect (i.e. invalid Swagger). Try changing the `security` value to the the following and it should recognize the multiple schemes:

```
"security": [
  { "username": [] },
  { "key": [] }
]
```

daniel-sullivan:
Sorry, I didn't word the issue correctly. I'm using the "pair of API keys" method as per https://swagger.io/docs/specification/authentication/api-keys.

On that spec they differentiate between:

**Pair of API keys**
```
security:
  - apiKey: []
    appId: []

```
**Two different API key auth mechanisms**
```
security:
  - apiKey: []
  - appId: []

```
Correct me if I'm wrong, (and the online YAML -> JSON convertor because I'm horrible with YAML...) but in my pair of keys being used together use case I should be using:

```
{
	"security": [
		{
			"apiKey": [],
			"appId": []
		}
	]
}
```

domaindrivendev:
Aha OK gotcha - that's actually a subtlety in the spec I wasn't aware of. Rswag currently assumes 1 security scheme per array item in the `security` field and furthermore it treats them as a logical AND (i.e. it tries to add both to the current request) when executing tests. Sound like this should be treated as logical OR and the syntax you've described should be for logical AND.

So def a bug, albeit a subtle one. Thanks for pointing it out

domaindrivendev:
This is fixed by ad9cd5de663967031d13e43e6297476363e3fc6a and published to rubygems (v1.5.2 and onward).

bsgreenb:
[...]

PR #89, opened 2017-08-03
Ability to use multiple apiKey entries in requests

daniel-sullivan:
As per issue #88


daniel-sullivan:
Please ignore this for the moment - forgot to branch and it's brought in other merges

PR #96, opened 2017-09-12
Adds request spec generator (fixed conflicts for PR #75)

sudoaza:
Seemed only fair that i should shamelessly stole this PR :)


BookOfGreg:
PR #75 was merged and tests were added.
Thanks for contributing this, the extra effort helps keep these things going.
Going to close.

PR #100, opened 2017-10-02
Support arel like query parameters

horiaradu:
* especially useful when combining with ransack

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -87,6 +87,8 @@ module Rswag
           "#{name}=#{value.join('|')}"
         when :multi
           value.map { |v| "#{name}=#{v}" }.join('&')
+        when :arel
+          value.map { |k, v| "#{name}[#{k.to_s}]=#{v}" }.join('&')
         else
           "#{name}=#{value.join(',')}" # csv is default
         end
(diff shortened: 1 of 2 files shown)


horiaradu:
@domaindrivendev could you take a look at this when you have time? Thanks!

domaindrivendev:
Hey there. Thanks for the contribution but my position at this point is to only support the `collectionFormat` values that are defined in the [Swagger specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md).

horiaradu:
What do you think about maybe using a different type, not ‘collectionFormat’. I think arel is a very useful type to supoort. I could adapt / change the code if you have an idea about what type we could use...

domaindrivendev:
Swagger is a technology agnostic specification for describing HTTP/REST based API's. Therefore, there should be nothing in the description that refers to Rails specific concepts. With that in mind, you're really just referring to a set of query parameters with a very specific naming format, which can be set up as follows:

```ruby
get 'Test' do
  tags 'Blogs'
  description 'test'
  operationId 'test'
  produces 'application/json'
  parameter name: :'test[foo]', :in => :query, :type => :string
  parameter name: :'test[bar]', :in => :query, :type => :string

  response '200', 'test' do
    let(:'test[foo]') { 'val1' }
    let(:'test[bar]') { 'val2' }
    run_test!
  end
end
```

horiaradu:
Ok, cool. Thanks a lot for the suggestion, I will use it.

Thanks for your time!

Issue #30, opened 2016-10-12
Fail to load swagger-ui assets


benoittgt:
Hello @domaindrivendev 

First thank you soooo much for the rewrite. I've edited my old request specs and it's work perfectly with query params. The json generated doesn't contain any errors when testing with http://editor.swagger.io/#/.

I have only one issue with swagger-ui.

When trying to go to `/api-docs` it gets lot's of missing assets like: 

```
ActionController::RoutingError (No route matches [GET] "/assets/swagger-ui/images/favicon-32x32.png")
```

Page looks like : 
<img width="233" alt="capture d ecran 2016-10-12 a 11 25 38" src="https://cloud.githubusercontent.com/assets/8417720/19304860/a7788bc2-906e-11e6-822d-e5ced61fe513.png">

also

``` sh
$ rails g rswag:ui:custom_ui                                                                             [11:26:20]
Running via Spring preloader in process 19510
Could not find generator 'rswag:ui:custom_ui'. Maybe you meant 'rswag:ui:custom', 'rswag:ui:install' or 'rswag:install'
Run `rails generate --help` for more options.
```

domaindrivendev:
Thanks for the quick feedback. Odd that the assets aren't working. Is this dev or prod? Do u def have asset pipeline enabled - can u access other assets in your app? Which version of Rails?

Re the generator, my bad I changed to "rswag:ui:custom" but have the old name in the docs. Will correct docs

benoittgt:
It's in dev env. Rails `4.2.5.2`
In dev env : 

```
  config.assets.compress = false
  config.assets.digest = false
```

No problem with the other assets of the app.

domaindrivendev:
Tried with Rails 4.2.0 and still can't repro. Do you have the following line in your routes.rb:

```
mount Rswag::Ui::Engine, at: 'api-docs'
```

Although I don't really think that should matter

domaindrivendev:
@benoittgt - can you try upgrading to 1.0.1 and let me know if it works?

benoittgt:
@domaindrivendev it's perfect now ! 👏🏻

unRARed:
Same problem in Rails 5 here with `config.api_only = true`. Also tried standardizing everything to `/swagger` with no change in results. Here's my current config:

* Gemfile: `gem 'rswag', '~> 1.1'` also tried `~> 1.0`
* Swagger doc: `./swagger/v1/swagger.json`  
[...]

PR #107, opened 2017-12-25
Adds API only mode support for rswag-ui

vigneshjayavel:
Fixes this 
#30


domaindrivendev:
The underlying issue should be now be fixed by #121 (available in version 2.0.0 and above). Please give it a go and let me know if it works

PR #126, opened 2018-05-25
Skip adding query params if their value is nil

pedantic-git:
This PR allows you to define query params as `nil` in your specs, and for the RequestFactory not to add them to the URL in this case.

You are still required to define all params with `let`, even if you define them as nil, and you can still add an empty string to the URL by defining them as `""`.

I know this is a slight change in functionality, but I'm sure this is what most people would mean when they say a query param is `nil` in a spec.

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -54,7 +54,7 @@ module Rswag
         definitions[key]
       end
 
-      def add_verb(request, metadata) 
+      def add_verb(request, metadata)
         request[:verb] = metadata[:operation][:verb]
       end
 
@@ -67,8 +67,11 @@ module Rswag
           end
 
           parameters.select { |p| p[:in] == :query }.each_with_index do |p, i|
-            template.concat(i == 0 ? '?' : '&')
-            template.concat(build_query_string_part(p, example.send(p[:name])))
+            value = example.send(p[:name])
+            unless value.nil?
+              template.concat(i == 0 ? '?' : '&')
+              template.concat(build_query_string_part(p, value))
+            end
           end
         end
       end
@@ -104,7 +107,7 @@ module Rswag
         end
 
         # Content-Type header
-        consumes = metadata[:operation][:consumes] || swagger_doc[:consumes] 
+        consumes = metadata[:operation][:consumes] || swagger_doc[:consumes]
         if consumes
           content_type = example.respond_to?(:'Content-Type') ? example.send(:'Content-Type') : consumes.first
           tuples << [ 'Content-Type', content_type ]
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -66,13 +66,12 @@ module Rswag
             template.gsub!("{#{p[:name]}}", example.send(p[:name]).to_s)
           end
 
-          parameters.select { |p| p[:in] == :query }.each_with_index do |p, i|
+          query = parameters.select { |p| p[:in] == :query }.map do |p|
             value = example.send(p[:name])
-            unless value.nil?
-              template.concat(i == 0 ? '?' : '&')
-              template.concat(build_query_string_part(p, value))
-            end
-          end
+            build_query_string_part(p, value) unless value.nil?
+          end.compact.join('&')
+          
+          template.concat('?' + query) unless query.empty?
         end
       end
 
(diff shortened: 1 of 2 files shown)


jsartin513:
Thanks for doing this! I ran into the same problems

@domaindrivendev any chance you have some time to review/merge this?

eric-hemasystems:
I wonder if we should really be requiring you to define params with `let` if it is not required. I noticed what seems to be an inconsistency in rswag.

If doing a `path` param it [explicitly sets `required` to true](https://github.com/domaindrivendev/rswag/blob/master/rswag-specs/lib/rswag/specs/example_group_helpers.rb#L39-L41). This implies to me that other parameter types are optional unless explicitly set to `true`.

But then when building the request [it checks if `required` is explicitly set to `false`](https://github.com/domaindrivendev/rswag/blob/master/rswag-specs/lib/rswag/specs/request_factory.rb#L36) implying that all params are required unless explicitly set to `false`.

My suggestion is that the first linked code is changed to:

```
attributes[:required] = if attributes[:in] && attributes[:in].to_sym == :query
  false
else
  true
end
```

This way parameters that are `query` are assumed to be optional and therefore the `let` can be omitted. I think optional query parameters is a reasonable default. Any other param type is assumed to be required and `let` is required.

Either way the generated Swagger file will always have the `required` field explicitly set so there is never vagueness on that.

pedantic-git:
@eric-hemasystems That does sound like a potentially better approach than mine. If you're happy to submit that as an alternative PR I'd be happy to close mine.

Although I think mine might still have some uses - for example, if you define a `context` in which you want to remove a parameter that was previously set with `let` in the parent context, so perhaps both should be implemented?

eric-hemasystems:
I'm happy to submit a PR although yours has been languishing since May and I hate to do the work only to have it get the same problem. If we could get some feedback from @domaindrivendev to get an indication regarding this I'm happy to do the legwork.

domaindrivendev:
@eric-hemasystems - you raise a good point, particulary as the [Swagger 2.0 spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-7) does indicate that the `required` field has a default of `false`.

However, rather than explictly setting the value behind the scenes, I'd prefer to keep the spec author in control of the generated Swagger document and just have the `request_factory.rb` interpret it correctly - i.e. treat non-path fields as optional unless the `required` field is explicitly provided and set. Does that make sense?

As it happens there's another tangental issue #145 which rightly calls out that it's impossible to test 4xx responses that are due to missing "required" parameters. So, I'm wondering if we can solve both issues by simply ignoring parameters unless a specific value has been provided via `let`.

That is, replace the following code (to identify applicable request parameters):

```ruby
(operation_params + path_item_params + security_params)
  .map { |p| p['$ref'] ? resolve_parameter(p['$ref'], swagger_doc) : p }
  .uniq { |p| p[:name] }
  .reject { |p| p[:required] == false && !example.respond_to?(p[:name]) }
```

With

```ruby
(operation_params + path_item_params + security_params)
  .map { |p| p['$ref'] ? resolve_parameter(p['$ref'], swagger_doc) : p }
  .uniq { |p| p[:name] }
  .reject { |p| !example.respond_to?(p[:name]) }
```

Interested to get peoples thoughts? Can you think of any down sides?

pedantic-git:
Personally I'd still like the flexibility of being able to remove query params by setting their value to `nil` - that  way you can have something like:

```ruby
response 200, 'blah' do
  let(:param) { 'normal' }
  it { ... }
  it { ... }
  context 'param is unset' do
    let(:param) { nil }
    it { ... }
  end
end
```

But it's not a *huge* advantage so I'm happy to drop this PR if you'd prefer the approach above as the only method for removing params.

domaindrivendev:
@pedantic-git - thanks for your feedback. I do see you're point from a convenience perspective. However, I also think that in an API context, there is a subtle but real difference between omitting a field entirely and providing it with a value of null. By conflating those two cases, rswag will constrain developers from creating tests that distingiush between the two.

pedantic-git:
@domaindrivendev That's certainly true about parameters of any type except `:query`, which is why my PR restricts this to `:query` params. In my opinion (you may disagree), the equivalent of `null` in a query param is to omit the param altogether.

But as I say - happy to go with the other implementation alone if you disagree!

rsailor:
I have to say I think whether "null" should unset a `let` definition would be very useful, but it's still a bit questionable, because setting something to null could have a lot of meanings. (But, @pedantic-git, you're right, "" and nil have the same effect for query params). Wish there was an Rspec method to unset `let`. 

In any case, I created a PR with @domaindrivendev 's solution to required parameters. I think it's a great idea, and wanted to get the ball rolling.

sergioisidoro:
According to this conversation, I will close this in favor of #236, since most parties have reacted positively to the last comment.

PR #143, opened 2018-08-23
Provide alternate rake task name to ensure test environment is loaded…

adamaig:
… with dotenv-rails.

New rake task name is consistent with how dotenv-rails detects spec running to set RAILS_ENV=test, https://github.com/bkeepers/dotenv/blob/a1fa729e5e6d96114f4fed9bcf19842a8e2776aa/lib/dotenv/rails.rb#L12

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,10 @@ Once you have an API that can describe itself in Swagger, you've opened the trea
 4. Generate the Swagger JSON file(s)
 
     ```ruby
-    rake rswag:specs:swaggerize
+    rake spec:swaggerize
     ```
+
+    (Deprecated) `rake rswag:specs:swaggerize`. If using `dotenv-rails` the test environment is not loaded unless `RAILS_ENV=test` is set.
 
 5. Spin up your app and check out the awesome, auto-generated docs at _/api-docs_!
 
--- a/rswag-specs/lib/tasks/rswag-specs_tasks.rake
+++ b/rswag-specs/lib/tasks/rswag-specs_tasks.rake
@@ -1,18 +1,22 @@
 require 'rspec/core/rake_task'
 
+namespace :spec do
+  desc 'Generate Swagger JSON files from integration specs'
+  RSpec::Core::RakeTask.new('swaggerize') do |t|
+    t.pattern = 'spec/requests/**/*_spec.rb, spec/api/**/*_spec.rb, spec/integration/**/*_spec.rb'
+
+    # NOTE: rspec 2.x support
+    if Rswag::Specs::RSPEC_VERSION > 2 && Rswag::Specs.config.swagger_dry_run
+      t.rspec_opts = [ '--format Rswag::Specs::SwaggerFormatter', '--dry-run', '--order defined' ]
+    else
+      t.rspec_opts = [ '--format Rswag::Specs::SwaggerFormatter', '--order defined' ]
+    end
+  end
+end
+
 namespace :rswag do
   namespace :specs do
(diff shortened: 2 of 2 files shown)


edusalguero:
Hi @adamaig,

I had a similar problem and I have fixed it using the proper environment, `RAILS_ENV=test` just as you said.

I have checked you PR but I can't understand how the new task can works. Are you sure that you have proposed a complete solution?

Regards

BookOfGreg:
I also don't understand why putting it in a different namespace would make a difference.

In general we could load the test bundler group within the rake task to avoid the `RAILS_ENV` requirement.

Going to close for now but will re-open for other solutions or an explaination.

adamaig:
@BookOfGreg @edusalguero Sorry for the very late reply. The problem I was suggesting a solution for comes from a combination of other software, specifically `dot-env` and `rspec-rails`, but it might manifest in other setups.

The discussion in `dot-env` and `rspec-rails` can be traced from the linked line of code, but the core of the issue is that the detection of specs being run is based on the task name starting with `spec`.

```
is_running_specs = Rake.application.top_level_tasks.grep(/^spec(:|$)/).any?
```

This convention is based on a similar task name convention used in rails https://github.com/rails/rails/blob/master/railties/lib/rails/test_unit/railtie.rb#L5

 If the intent is to have the swagger task run under the spec setup space with the `test` env settings by default, then this change made sense to me. Otherwise, someone may run the rswag task against the development environment, which would behave differently than when running specs, and might have destructive results.

BookOfGreg:
I'd be interested in finding out if there's a way of just running rspec with all your current config with an additional flag, that would be the ideal situation of whatever your rspec does hooking into that. I don't really want to have to cater to any possible combo of gems in a specific way because that seems like an endless task.

Reopened as you still have interest in it but I'd like to look for a generic approach to replacing the line you singled out above.

adamaig:
Fair-enough. I'm wondering if you'd be open to treating the new task name as an addition/alternative invocation of the `rswag:specs` task, rather than as a deprecated option as originally suggested in this PR.

Adding `RAILS_ENV=test` to the cli achieves the expected outcome, but this means that the default behavior isn't what anyone expects if they are using `dot-env`. Given `dot-env` usage seems fairly common, it might bear at least a comment on the issue in the README if the task addition isn't acceptable.

I think the generic solution would be to move the task under the `spec` namespace because that is the convention for running spec-related tasks. The reason I pointed out a similar line in the rails core for `test` was to point at the foundations for that design choice. Because that seems to be the broader community convention, I think it helps normalize the rswag tasks to the community expectations, which might avoid the issue of endless gem-specific modifications for rswag.

I'll make the changes to list both task names with an explanation if you're open to that option.

sergioisidoro:
@adamaig are you still interested in pushing this forward? Is this still relevant?

Personally I think that you have a valid argument that generation of the docs could be considered a `spec` namespace task, since I think I've also encountered this issue before, and realising I needed to explicitly add `RAILS_ENV=test`.

Also, I think `swaggerize` is unique enough not to colide with other gems that might add to the `spec` namespace, so this would be ok, although `spec:rswag:swaggerize` could also be a possibility.

sergioisidoro:
Re-ping @adamaig are you interested in working on this?
Otherwise we will close given how old this PR is

adamaig:
@sergioisidoro I believe this is still an issue. I'll take a stab at rebasing this work on the latest mainline HEAD and confirm this is still an issue this week.

PR #152, opened 2018-09-21
Call disable_monkey_patching! for RSpec configuration, to fix IRB context warning in Rails console

### LINKED ISSUE #151: irb: warn: can't alias context from irb_context
I got this warning whenever starting the Rails console, and I tracked it down to the `rswag` gem.
It was also caused by other gems in the past, all related to Rspec: 

* rspec-rails: https://github.com/rspec/rspec-rails/issues/1645
* guard-rspec: https://github.com/guard/guard-rspec/issues/396
* paper_trail: https://github.com/paper-trail-gem/paper_trail/commit/d048508f38f5e6acaf244869620ca7076f4e3cf7

It would be great to fix this warning in `rswag`.

ndbroadbent:
Fixes the IRB warning in #151.

This disables the RSpec monkey patching. It means that you need to change any top-level `describe` to `RSpec.describe`, but it should only be one change at the top of each spec file. You can continue to use `describe` and `context` within the block.

This will affect the specs of any Rails application that uses the `rswag` gem, but I think this is a best-practice now, and it's good to encourage people to use the `disable_monkey_patching!` option.

### DIFF
--- a/rswag-specs/spec/generators/rswag/specs/install_generator_spec.rb
+++ b/rswag-specs/spec/generators/rswag/specs/install_generator_spec.rb
@@ -4,7 +4,7 @@ require 'generators/rswag/specs/install/install_generator'
 module Rswag
   module Specs
 
-    describe InstallGenerator do
+    RSpec.describe InstallGenerator do
       include GeneratorSpec::TestCase
       destination File.expand_path('../tmp', __FILE__)
 
--- a/rswag-specs/spec/rswag/specs/configuration_spec.rb
+++ b/rswag-specs/spec/rswag/specs/configuration_spec.rb
@@ -3,7 +3,7 @@ require 'rswag/specs/configuration'
 module Rswag
   module Specs
 
-    describe Configuration do
+    RSpec.describe Configuration do
       subject { described_class.new(rspec_config) }
 
       let(:rspec_config) { OpenStruct.new(swagger_root: swagger_root, swagger_docs: swagger_docs) }
--- a/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
+++ b/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
@@ -3,7 +3,7 @@ require 'rswag/specs/example_group_helpers'
 module Rswag
   module Specs
 
-    describe ExampleGroupHelpers do
+    RSpec.describe ExampleGroupHelpers do
       subject { double('example_group') }
 
       before do
--- a/rswag-specs/spec/rswag/specs/example_helpers_spec.rb
+++ b/rswag-specs/spec/rswag/specs/example_helpers_spec.rb
@@ -3,7 +3,7 @@ require 'rswag/specs/example_helpers'
 module Rswag
   module Specs
 
-    describe ExampleHelpers do
+    RSpec.describe ExampleHelpers do
       subject { double('example') }
 
       before do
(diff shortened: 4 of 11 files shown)


olleolleolle (approved):
This improves the developer experience, **and** helps RSpec use less magic.

rswag-specs/lib/rswag/specs.rb
```diff
@@ -9,6 +9,10 @@ module Specs
 
     # Extend RSpec with a swagger-based DSL
     ::RSpec.configure do |c|
+      if c.respond_to?(:disable_monkey_patching!)
+        # Fix "irb: warn: can't alias context from irb_context." in Rails console
+        c.disable_monkey_patching!
```
BookOfGreg: We can be compliant with if someone has the `disable_monkey_patching!` option but it is not our place to set this setting on other project's RSpec config.

This suggestion comes from a good place that I agree with, so the best we could do is leave this commented as a suggestion.

BookOfGreg (changes requested):
Thank you for making the tests compliant, I will accept those in and the new tests for the generator come with `RSpec.` infront of describes in the generator.
We won't set options for other parts of RSpec though as that choice is outside of our context. Once that's removed or commented I can accept this.

rswag-specs/lib/rswag/specs.rb
ndbroadbent: Ah that's a good point, I have removed the call to `disable_monkey_patching!`, so this PR just adds `RSpec` in front of the `describe` calls. That still seems to fix the warnings for me.

BookOfGreg (approved):

rswag-specs/lib/rswag/specs.rb
BookOfGreg: I think I just made a silly mistake. For some reason I assumed this was the spec_helper template that was copied into the project. This is for just fixing this projects specs, I was wrong to ask for this to be removed, and we could add a comment to the template that is copied in suggesting it.

That's what I get for merging things at midnight 😆

BookOfGreg:
Thank you for contributing this to the project, helping make the tests better is a big assist for maintaining this. 🎉

ndbroadbent:
Great, thank you! 

But sorry, I was actually just about to comment to say that I was wrong! Because calling `c.disable_monkey_patching!` is actually the only way to silence this warning: `irb: warn: can't alias context from irb_context.` (Even though I'm calling this in my `spec_helper.rb`, this doesn't get loaded when Rswag is loaded in the development environment.)

I've realized that I can fix this in my own application by calling it at the top of my Rswag initializer, in `config/initializers/rswag.rb`:

```
# frozen_string_literal: true

# Silence "irb: warn: can't alias context from irb_context." warning
RSpec.configure(&:disable_monkey_patching!)

# ...
```

I'm not sure if there's a better way to do that, but this is working for me!

BookOfGreg:
I realise I was wrong above asking you to remove it, does adding the disable monkey patching back in to the spec file fix the issue? If so I'll accept and merge another PR bringing that back in.

ndbroadbent:
Hello, yes if monkey patching is disabled then the warning goes away. But I also realized that it's not ideal to load RSpec in the development environment. So it would be even better if rswag could load different things during development, and only load RSpec when it's running the tests. I'm not too sure how that could be achieved, but that would be another approach. (The warning doesn't appear at all if RSpec hasn't been required.)

PR #177, opened 2018-12-22
Show response body when the response code is unexpected. 

ndbroadbent:
This makes it much easier to debug test failures.

Instead of just:

```
Expected response code '400' to match '201'
```

It will show the response body as well, so you can see any errors. (I always have to manually add code that prints out the response body so that I can see what happened.)


jaydorsey:
This would be a really nice feature. Adding a configuration-level flag might be useful too, with default behavior to not show test failures so the current behavior is maintained but developers can opt in to it

chdezmar:
Any plans to merge this soon? I think this is a super helpful feature, and I would say even essential.

jaydorsey:
I've been using a fork that has these changes rolled in and they're invaluable for debugging when you have an error.

PR #180, opened 2019-01-07
Add a 'rake rswag' that runs swaggerize as the default

pedantic-git:
What do you think of this idea? I find myself running `rake rswag:specs:swaggerize` very frequently, and it's quite a mouthful for such a common command. I think others could benefit from it being aliased to `rake rswag`.


BookOfGreg:
Agreed :+1:

PR #187, opened 2019-01-30
allow headers to be set in the configuration of rswag-api

pjdavis:
Allows the headers for serving from the RSwag::API engine to be set in the configuration initializer. 

Merges with the existing 'Content-Type' header, so that will always be set.


BookOfGreg:
This is a good PR. We do need to consider now that the RSwag API could serve both swagger.json and swagger.yml or swagger.yaml files.
Do you have any thoughts on how this PR could be modified to accomodate that need?

I've been looking lately at [Rack::Static](https://github.com/rack/rack/blob/master/lib/rack/static.rb) and Rack::File which do have built-in file handlers with some level of configuration for customization.
I wonder if instead of rolling our own we can make use of those.

BookOfGreg (approved):

BookOfGreg:
Thank you  for writing tests along with the code. Most of the others that I could merge are stuck because of a lack of tests.

PR #190, opened 2019-02-06
Allow parsing of yml swagger files in rswag-api

bclewis1:
Hello,
This project is really nice, I enjoyed writing specs using the dsl and everything felt very natural.
We are currently monkey patching our project as below because we have a large existing collection of swagger yamls.


rswag-api/lib/rswag/api/middleware.rb
```diff
@@ -29,6 +30,18 @@ def call(env)
 
       private
 
+      def parse_file(filename)
+        if /\.yml$/ === filename
```
jaydorsey: Two suggestions: 

- `filename.end_with?('.yml')` (if you only want .yml)
- `/\.ya?ml$/ === filename` (the [yaml faq](https://yaml.org/faq.html) suggests .yaml, so might be nice to support it)

jaydorsey:
I like this as an idea. Even the swagger tool online prompts you to convert your json to yaml when you paste it.  Plus, most of the examples on the swagger site are in yaml.

rswag-api/lib/rswag/api/middleware.rb
bclewis1: :+1:, sorry about the delay fixing this

BookOfGreg:
I just spent a couple of days learning how Rack serves files statically without needing to deal with mimetypes without any risk of path traversal issues:
https://github.com/rack/rack/blob/master/lib/rack/static.rb

It's quite interesting.

The takeaway from this is I'll merge this PR (yay! YAML support) and you may see me working on this set of files in the near future.
Also if you like please look into how static file serving appears to be done by others as I feel that I need certainty that this solution doesn't put anyone at risk.

Thank you for your contributions!

PR #197, opened 2019-02-18
Allow optional parameter named 'include'

### LINKED ISSUE #188: Parameter names matching rspec built-in matchers are always required
Given a parameter defined like this:

    parameter name: :include, in: :query, type: :string, required: false

When running my specs, I get the following error:

```
URI::InvalidURIError:
bad URI(is not URI?): 
http://www.example.com:80/users/f86316df-6892-439f-af9e-ba15ba731c21?include=#<RSpec::Matchers::BuiltIn::Include:0x00007fd507afeb80>
```

Even though required is set to false, it still expects a value. It appears there is a conflict with the RSpec `include` matcher & the parameter name being `include`. 

I can overcome this error two ways:

- By defining `let(:include) {''}` in my `response` block (this is what I'm doing)
- Changing my parameter name (can't do this, because we're using jsonapi)

jaydorsey:
A parameter named include was creating an issue, because even when not
required, it wasn't being rejected because the `respond_to?` check was
always returning `true`. An `example` was responding to the built-in
RSpec method `include`, versus looking for the memoized `let` variable.

This closes #188 

I think I'd ideally like to have checked this another way, but I don't see a way to
differentiate (on an example) whether a method is defined via a `let` or it's a
method that the example responds to

The reason behind this PR is because the `include` keyword as a query parameter
is a jsonapi standard convention, so I can't well change the parameter name and
follow the standard.

There is a workaround that kind-of works, setting a `let(:include) { {} }` before each
test; but that results in a URL with an empty parameter and isn't intuitive, or
correct. 

This PR allows us to generate documentation without having to specify the
optional `include` query parameter

Update: I need to look into this in depth, again. This still breaks the `include` matcher
when you use a `let(:include)`. I think that's an rspec issue though and I'm not sure 
that's something we can overcome

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -33,7 +33,11 @@ module Rswag
         (operation_params + path_item_params + security_params)
           .map { |p| p['$ref'] ? resolve_parameter(p['$ref'], swagger_doc) : p }
           .uniq { |p| p[:name] }
-          .reject { |p| p[:required] == false && !example.respond_to?(p[:name]) }
+          .reject { |p|
+            p[:required] == false &&
+            (!example.respond_to?(p[:name])) ||
+            (example.respond_to?(p[:name]) && example.send(p[:name]).is_a?(RSpec::Matchers::BuiltIn::Include))
+          }
       end
 
       def derive_security_params(metadata, swagger_doc)
--- a/rswag-specs/spec/rswag/specs/request_factory_spec.rb
+++ b/rswag-specs/spec/rswag/specs/request_factory_spec.rb
@@ -127,6 +127,34 @@ module Rswag
           end
         end
 
+        context 'optional parameter named include' do
+          context 'not included' do
+            before do
+              metadata[:operation][:parameters] = [
+                { name: 'include', in: :query, type: :string, required: false }
+              ]
+              allow(example).to receive(:include).and_return(RSpec::Matchers::BuiltIn::Include.new)
+            end
+
+            it 'builds request hash without them' do
+              expect(request[:path]).to eq('/blogs')
+            end
+          end
+
+          context 'included' do
+            before do
+              metadata[:operation][:parameters] = [
+                { name: 'include', in: :query, type: :string, required: false }
+              ]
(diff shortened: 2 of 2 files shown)


yxhuvud:
This would be extremely nice to have considering that `include` is one of the keywords of `json:api`. See https://jsonapi.org/format/#fetching-includes

jaydorsey:
I've since discovered there's still an issue with how this all works, and I'm not sure if it's rspec or rswag side of things:

```
let!(:include) { 'foo' } # ?include=foo

run_test! do
  # This include breaks, because it gets redefined
  expect(<parsed response>).to include([1, 2, 3])
end
```

WIth RSpec, the include matcher appears to get overwritten by the `let`, because the `let` is defined at the top level. I can live with this, however.

My work around is to just never use the `include` matcher when I'm doing a test that uses an `include` param (I'll do `eq([1, 2, 3])` for example) but it's not very elegant

This PR will still let you, at least, use `include` as a parameter name so I think it's still a relevant change. Without this PR the specs always break:

Here's an example of one of my parameters with an `include` defined, to document it:

     parameter name: :include, in: :query, type: :string, required: false

And a sample spec (I'm not even using include in the querystring, or testing it):

        context 'default' do
          let!(:task0) { create(:task) }

          run_test! do
            expect(json_response).to match_json_schema('response')
          end
        end

And here's a sample error:

```
  1) tasks /tasks get tasks default returns a 200 response
     Failure/Error: raise InvalidURIError, "bad URI(is not URI?): #{uri.inspect}"

     URI::InvalidURIError:
       bad URI(is not URI?): "http://www.example.com:80/tasks?page[number]=1&page[size]=25&include=#<RSpec::Matchers::BuiltIn::Include:0x00007f97d56ab8b8>"
```

This PR fixes that error

jaydorsey:
Closing this PR. Someone else can feel free to re-open.

I no longer use rswag and don't have the time to investigate a better solution for this. Hopefully the example I provided will be useful to others.

kwent:
This should be reopened this is still an issue !

varg90:
@kwent unfortunately this PR was not merged, but there is another PR https://github.com/rswag/rswag/pull/436
...which is not merged also.
If there is another way to use `parameter :'include', ...` in our rswag request tests please let us know.

PR #209, opened 2019-04-17
Gemfile: Use exact sqlite3 versions to fix the build

olleolleolle:
This PR is about **fixing the build failure** re: exact version of `sqlite3` used by Rails.

  - Avoids `Gem::LoadError` about sqlite3 1.3.6 needs

See https://stackoverflow.com/a/54729071/267348 for more details about why this happens.


BookOfGreg:
This has now been updated and Travis builds on master.
Thanks for your contribution!

PR #214, opened 2019-05-02
Support custom properties to allow OpenAPI extensions

bubbavcita:
Support custom properties to allow OpenAPI extensions such as Amazon API Gateway.


hat-tinypulse:
👍

Issue #178, opened 2018-12-28
Is it possible to document a 500 error response in swagger?


arianf:
Is it possible to document a 500 error response in swagger?

I want my swagger docs to show what JSON structure will be passed when a 500 or other error occurs, but I'm unable to trigger the error in the Rspec test. Is it possible to have swagger document a method without running `run_test!` within that block?

for example, I may have this defined
```ruby
response '500', 'internal server error' do
  schema type: :object,
         properties: {
           message: { type: :string }
         }

  run_test!
end
```

when I run Rspec, I get an failure, if I knew how to trigger a 500, I would fix it. But that doesn't mean my API might never respond with such a response...

```
     Rswag::Specs::UnexpectedResponse:
       Expected response code '200' to match '500'
```

But when I generate docs... it works fine.
```bash
rake rswag:specs:swaggerize
```

<img width="459" alt="screen shot 2018-12-27 at 11 00 30 pm" src="https://user-images.githubusercontent.com/7397857/50506021-89021f80-0a2b-11e9-897f-d7887be40cf1.png">

Now my question is, is there a way to make Rspec ignore actually trying to test a spec for rswag?

like `xresponse` instead of `response`, or is there a way to remove `run_test!` while having rake swaggerize still look at those blocks? Maybe a `run_test` (without the !) to signify the test isn't passing, but you still want it to be included in the swagger docs?

This could apply to any error code that you might not be able to trigger, but you know if it does happen, this would be the response. How do you document that in rswag?

ToniTornado:
Same problem here. I made my tests pass by manually triggering a `500` response if a certain condition is met. I'm authorizing every request with a token; in my tests I set this token to "500" (otherwise an invalid token) and check for that in the beginning of my authorize `before_action`.

my_spec.rb:
```ruby
response '500', 'internal server error' do
  let('X-API-TOKEN') { '500' }
  run_test!
end
```
base_controller.rb:
```ruby
def authorize
  api_token = request.headers['X-API-TOKEN']
  if api_token == '500'
    head 500
    return
  end
  # Check token here...
end
```

I like your `xresponse` solution to document status that I'm unable to trigger.

rheasunshine:
Was there ever a solution for this?

rheasunshine:
I put a new method in my swagger_helper.rb and call it instead of run_test! to accomplish this.

```
def document_response_without_test!
  before do |example|
    submit_request(example.metadata)
  end

  it 'adds documentation without testing the response' do |example|
[...]

PR #219, opened 2019-06-20
[Issue #178] Add support for documenting an error response

rheasunshine:
Solution for https://github.com/domaindrivendev/rswag/issues/178

Add support for documenting a response without validating it as in run_test!

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -97,6 +97,29 @@ module Rswag
           end
         end
       end
+
+      def document_response_without_test!
+        # NOTE: rspec 2.x support
+        if RSPEC_VERSION < 3
+          before do
+            submit_request(example.metadata)
+          end
+
+          it 'adds documentation without testing the response' do
+            # Only check that the response is present
+            expect(example.metadata[:response]).to be_present
+          end
+        else
+          before do |example|
+            submit_request(example.metadata)
+          end
+
+          it 'adds documentation without testing the response' do |example|
+            # Only check that the response is present
+            expect(example.metadata[:response]).to be_present
+          end
+        end
+      end
     end
   end
 end
--- a/test-app/spec/integration/blogs_spec.rb
+++ b/test-app/spec/integration/blogs_spec.rb
@@ -27,6 +27,13 @@ RSpec.describe 'Blogs API', type: :request, swagger_doc: 'v1/swagger.json' do
           expect(response.body).to include("can't be blank")
         end
       end
+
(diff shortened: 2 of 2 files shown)


sergioisidoro:
@rheasunshine would it be possible to add this to the documentation (one liner would be enough in the readme)?

I'm wondering if there would be a realistic way of adding a test runner (eg `run_test_with_error!`) that would actually raise an error in the controllers during the execution of the request, to actually test instead of just adding to the documentation. Did you explore that possibility?

jtannas:
+1 on Sergio's thoughts. I fiddled around a bit and here's what I came up with using existing tools in RSwag:
```ruby
response '502', 'Bad Gateway' do
  let(:Authorization) { external_api_token }

  schema(
    type: :object,
    properties: {
      status: { type: :string, enum: ['502'] },
      message: { type: :string, enum: ['Bad Gateway'] },
      description: { type: :string },
    },
  )

  before do |example|
    allow_any_instance_of(ExampleController).to receive(:create).and_raise BadGateway
    suppress(BadGateway) { submit_request(example.metadata) }
  end

  it 'returns an appropriate error description JSON' do |example|
    assert_response_matches_metadata(example.metadata)
  end
end
 ```

The hiccup here is that stubbing `ExampleController` to throw a 500 error requires knowledge of your application code, so I don't think it's something we can roll directly into RSwag. I could imagine adding `run_test_with_error!(ExpectedException)` to simplify it into this though:

```ruby
response '502', 'Bad Gateway' do
  let(:Authorization) { external_api_token }

  schema(
    type: :object,
    properties: {
      status: { type: :string, enum: ['502'] },
      message: { type: :string, enum: ['Bad Gateway'] },
      description: { type: :string },
    },
  )

  before { |example| allow_any_instance_of(ExampleController).to receive(:create).and_raise BadGateway }

  run_test_with_error!(BadGateway)
end
 ```
 
 In the meantime, I think the 1st version  could go into an FAQ alongside https://github.com/rswag/rswag/issues/140#issuecomment-449047723

**Comment by jtannas** [comment]
Coming back to this one sorting thorough issues and PRs.
After looking at it a second time, it's a bit weird that a new helper or a strange `xit` hack is needed to document a response without having any examples :thinking: 

As a user of the tool, I would expect to just be able to call `response` without a block and have it show up in the documentation. 
```ruby
response '500', 'Internal Server Error'
```

It's probably more complicated to implement on our end but it's one less thing the user has to learn. Theoretically [the current wiki solution](https://github.com/rswag/rswag/wiki/Examples#documenting-a-5xx-error-response) works and adds testing value but it's brittle and takes a fair amount of effort.

PR #235, opened 2019-10-02
Update contributing instructions. Add changelog

BookOfGreg:
Two changes:
- Update Contributing instructions.
- Add Changelog.

Contributing instructions now leverage the ci files, if the ci files are wrong then travis isn't building anyway so this should be a cheap way of guaranteeing everyone can get started.
Changelog may want to be backfilled at least with major changes between 2.x and last 1.x.

Edit:
I'd also like to add instructions for skipping the travis build (they provide resources for free, good to know how not to waste them), and also other common GitHub docs such as Issue and PR templates, probably a security.md also.

### DIFF
--- a/test-app/db/schema.rb
+++ b/test-app/db/schema.rb
@@ -2,15 +2,15 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
-# you'll amass, the slower it'll run and the greater likelihood for issues).
+# This file is the source Rails uses to define your schema when running `rails
+# db:schema:load`. When creating a new database, `rails db:schema:load` tends to
+# be faster and is potentially less error prone than running all of your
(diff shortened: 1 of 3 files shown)


test-app/db/schema.rb
```diff
@@ -2,15 +2,15 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
-# you'll amass, the slower it'll run and the greater likelihood for issues).
+# This file is the source Rails uses to define your schema when running `rails
```
BookOfGreg, line 5: This change is just Rails 6 changing this text. I'm committing it in on the assumption that most people will be wanting to run the newest rails.
Let me know if this causes anyone running Rails 5 or lower for testing. I'll double-check myself before merging.

BookOfGreg:
May I ask, @domaindrivendev do you prefer many small releases or a few larger ones? I could release what is currently in master as 2.0.6 and then do a 2.0.7 with updated swagger UI. Or do you prefer them both rolled into the next teeny release?

If you have no preference I'll probably do many teeny releases as I quite like continuously releasing.

Issue #145, opened 2018-08-29
Exclude params in a test


jsartin513:
Thanks for this gem, it's been a huge help to us!

We can't test a lot of our 400-status use cases because we're testing the case of a missing parameter. If we don't define the parameter, we get a `NoMethodError` trying to run the test. 
We have in some cases worked around it by setting the parameter to nil, but that's not testing the case that we're looking for.

Is there a way to do this? Or would this be something you might add in the future?

kubi-dev:
> We have in some cases worked around it by setting the parameter to nil, but that's not testing the case that we're looking for.

It is fine, you should be happy with this solution

jaydorsey:
@jsartin513 can you post an example of one of your specs? I found that setting `required: false` on my parameter definitions allowed me to run the tests & generate the documentation without having to pass in the parameter as nil (or assign a value).

PR #236, opened 2019-10-03
Parameters not required by default

rsailor:
The Swagger (2.0) spec states that without specifying the required property, parameters are not required by default. Rswag throws a NoMethodError if a parameter is not explicitly required and also not defined in an example. In effect it forces non-required parameters to be defined before the spec runs.

I shamelessly took @domaindrivendev's solution from [this comment](https://github.com/rswag/rswag/pull/126#issuecomment-432290716) (which I tested and find to be the best solution), wrote some tests, an created this PR in the hope we can get this merged and avoid the confusion this bug causes, in the future.

To note, this solution means it's up to the application developer to test for missing required parameters via their API. But, parameters should only be required if the API responds with an error, anyway. (see [this issue](https://github.com/rswag/rswag/issues/145))

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -1,6 +1,7 @@
 require 'active_support/core_ext/hash/slice'
 require 'active_support/core_ext/hash/conversions'
 require 'json'
+require 'byebug'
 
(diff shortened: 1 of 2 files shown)


sergioisidoro:
@rsailor Is this still relevante since we're already in OAS3?

PR #237, opened 2019-10-04
Add options and trace verbs

### LINKED ISSUE #231: Support OPTIONS verb
Having a peek into the code, I see the verbs `[ :get, :post, :patch, :put, :delete, :head ]` are supported, but no [OPTIONS](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html). I think it would be good to be able to write calls that respond to the options verb.

This isn't as esoteric of a feature request as it may appear, I think. I discovered a potential opportunity to make use of it recently while working on a moderately complex web service method for querying records, where it would benefit from a counterpart OPTIONS verb call to retrieve a (albeit, mostly static) list of query keys. I guess I could do this as a separate GET call, but you know, being able to cease an available rfc to do the job would be nice.

PMGH:
### Related Issues

Fixes https://github.com/rswag/rswag/issues/231

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -7,7 +7,7 @@ module Rswag
         describe(template, metadata, &block)
       end
 
-      [ :get, :post, :patch, :put, :delete, :head ].each do |verb|
+      [ :get, :post, :patch, :put, :delete, :head, :options, :trace ].each do |verb|
         define_method(verb) do |summary, &block|
           api_metadata = { operation: { verb: verb, summary: summary } }
           describe(verb, api_metadata, &block)
--- a/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
+++ b/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
@@ -24,7 +24,7 @@ module Rswag
         end
       end
 
-      describe '#get|post|patch|put|delete|head(verb, summary)' do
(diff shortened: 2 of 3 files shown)


BookOfGreg:
This does seem to input the verb as metadata as seen: `rswag/rswag-specs/lib/rswag/specs/example_helpers.rb @ line 20`
```
pry(#<RSpec::ExampleGroups::ResourceAPI::ApiResourcesId::Options::OptionsForResourceRetrieved>)> request
=> {:verb=>:options,
 :path=>"/api/resource/715b26bc-77c1-4eeb-ace3-d2e1e2be19e2",
 :headers=>{"HTTP_ACCEPT"=>"application/json"}}
```
However it does crash at this point as `Options` is undefined in ActionDispatch::IntegrationTest.

This may be a limitation in Rails, I wonder if we could lean on some example of Rails returning Options requests some other way.

BookOfGreg:
With the following two modifications to Rails itself I was able to generate this on your PR:
https://github.com/rails/rails/pull/37370
https://github.com/rails/rails/pull/37369

Screenshot 2019-10-04 at 21 11 49

BookOfGreg:
With the above 2 tweaks to Rails merged, I feel comfortable to merge this also.
For anyone visiting this thread in the future, you will need to be using an up to date version of rails or another means of actually providing the options verb, but rswag will be able to generate it once it is available. Trace hasn't been included in the patch to Rails.

We may want to add some form of documentation for people attempting to use Options and Trace based on what they're likely to see and how to fix it.

PR #238, opened 2019-10-05
add support reusing responses

zgid123:
Hi all,

First of all, thanks for creating this awesome gem and sorry for my bad English.

I tried to use [**Reusing Responses**](https://swagger.io/docs/specification/2-0/describing-responses/), but it seems this current version does not support it.

I tried to check your code and couldn't figure it out, so if the current version already supports, please show me how to do it.

Here the example config:

I define responses in _swagger_helper_ like this

```
{
  responses: {
    Unauthorized: {
      description: 'Unauthorized',
      schema: {
        '$ref': '#/definitions/Error'
      }
    }
  }
}
```

I want to generate config response for my api endpoint in _swagger.json_ like this

```
"401": {
  "description": "",
  "$ref": "#/responses/Unauthorized"
}
```

With this **ref** method, I will do like this

```
response '401', '' do
  ref('#/responses/Unauthorized')
  run_test!
end
```

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -58,6 +58,10 @@ module Rswag
         metadata[:response][:schema] = value
       end
 
+      def ref(value)
+        metadata[:response][:$ref] = value
+      end
+
       def header(name, attributes)
         metadata[:response][:headers] ||= {}
         metadata[:response][:headers][name] = attributes
--- a/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
+++ b/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
@@ -150,6 +150,15 @@ module Rswag
         end
       end
 
+      describe '#ref(value)' do
+        before { subject.ref('#/responses/Unauthorized') }
+        let(:api_metadata) { { response: {} } }
+
+        it "adds to the 'response' metadata" do
+          expect(api_metadata[:response][:$ref]).to match('#/responses/Unauthorized')
+        end
+      end
+
       describe '#header(name, attributes)' do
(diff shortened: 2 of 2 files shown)


andtherand:
I know this is kind of an old PR but as far as I know and understand the OpenAPI Spec V3 the things one can reference with the $ref all have to live under the key components.

The structure therefor needs to look like this:

```json
// your awesome endpoints 
// ...
"components": {
  "responses": {
    // your reusable response
  }
}
```
The corresponding $ref would then look like this: `$ref: '#/components/responses/YOUR_RESPONSE' ``

https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.0.md#componentsObject

masterkain:
I spent a lot of time trying to make `components/responses` work, then I saw this pr.

what's the status about this feature? should it work out of the box currently? I get `The fragment '/components/responses' does not exist on schema 20207978-375e-514a-bbbe-243e6211ae4a`

Envek:
Please reopen this pull request, it is a good one. Hope to see it merged.

I made a following monkey patch based on this pull request as a guidance and it works on v2.15.0 without any changes:

```ruby
module Rswag
  module Specs
    module ExampleGroupHelpers
      def ref(value)
        metadata[:response][:$ref] = value
      end
    end
  end
end
```

I use it as follows:

```ruby
response "403", "Forbidden" do
  ref "#/components/responses/Forbidden"
  run_test!
end
```

And it generates me as expected:
```yaml
responses:
  '403':
    description: Forbidden
    "$ref": "#/components/responses/Forbidden"
```

Note that description and other options are explicitly allowed to be specified along `$ref` since OAS 3.1, see https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#no-ref-siblings

Issue #215, opened 2019-05-13
rswag:specs:swaggerize not found


victorhazbun:
The command `rake rswag:specs:swaggerize` is not available with using the following configuration from the README:

```ruby
# Gemfile
gem 'rswag-api'
gem 'rswag-ui'

group :test do
  gem 'rspec-rails'
  gem 'rswag-specs'
end
```

It only works when using:

```ruby
gem 'rswag'
```

joshling1919:
This might be because `rswag:specs` is only inside of your test group.
try `RAILS_ENV=test rake rswag:specs:swaggerize`

Issue #200, opened 2019-02-22
Rake Task: rswag:specs:swaggerize not available


nicwillemse:
For some reason im not able to execute the swaggerize rake task, I have tried to reinstall this did not work

Environment Details :
```
Using rswag-api 2.0.5
Using rswag-specs 2.0.5
Using rswag-ui 2.0.5
```
Error: 

```
$ bundle exec rake rswag:specs:swaggerize
rake aborted!
Don't know how to build task 'rswag:specs:swaggerize' (See the list of available tasks with `rake --tasks`)
/Users/nic/.rbenv/versions/2.5.3/bin/bundle:23:in `load'
/Users/nic/.rbenv/versions/2.5.3/bin/bundle:23:in `<main>'
(See full trace by running task with --trace)
```

Petercopter:
@nicwillemse Same problem here. I see it in the list of tasks, but I'm unable to run it... did you resolve?

nicwillemse:
hey @Petercopter , I ended up redoing the configuration and reinstalling which fixed the issue. 

So unfortunately I did not get to pin point the actual problem.

Petercopter:
@nicwillemse Thanks for the response! I figured you closed the issue because it was something you were able to resolve by messing with it. I will do the same.

ealvaro:
@nicwillemse and @Petercopter
You gotta execute the rake within the `test` environment as below:
`rails rswag:specs:swaggerize RAILS_ENV=test `
if you avoided loading rspec in other bundler groups like development or production.

PR #247, opened 2019-10-17
Update README

laurawatson:
## Problem
Unable to execute the swaggerize task without specifying the test environment. 

## Solution 

Advise people on how to avoid this common problem. 

### Relevant issues 
https://github.com/rswag/rswag/issues/215
https://github.com/rswag/rswag/issues/200

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,14 +26,15 @@ Once you have an API that can describe itself in Swagger, you've opened the trea
     gem 'rswag'
     ```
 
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Ah you're referring to the common `RAILS_ENV=test rake rswag:specs:swaggerize` thing.

Yep, I think this advice will help some casual users.

PR #252, opened 2019-11-02
Use a prebuilt ruby

BookOfGreg:
# Problem
Runing our Build and Test phases takes only 15s but a build takes 5 minutes.
https://travis-ci.org/rswag/rswag/jobs/606449612?utm_medium=notification&utm_source=github_status

Ubuntu 16.04 Xenial only has ruby 2.5.x prebuilt.
This meant installing ruby took 1/3 of the total time.

# Solution
Upgrade to Ubuntu 18.04 Bionic
https://docs.travis-ci.com/user/reference/bionic/
Pre-installed Rubies: 2.3.8, 2.4.5, 2.5.3 and 2.6.3.


BookOfGreg:
After changes it looks like it takes 38s for the visible bits of travis, of which ~17 is an apt-get update, about 30s for travis to boot.
I've knocked about 20s off the run per stage so far.

BookOfGreg:
Using conditional builds makes this all much faster. We will see if it works properly next deploy.

PR #254, opened 2019-11-15
Parameterize the pattern for test locations

karljohansson85:
This commit allows users to specify search patterns when finding tests to
swaggerize. Omitting the pattern parameter makes rswag search with the default
patterns.

A typical usecase for this feature is when you already have a test suite set up
and you want to use rswag for generating swagger docs rather than high-coverage
testing.

Usage:

rake rswag:specs:swaggerize PATTERN='/your_path..'


BookOfGreg:
Looks good, thanks for contributing this :)

PR #255, opened 2019-11-15
Allow tests to be run without generating docs

karljohansson85:
By providing the 'document: false' metadata, tests will be run but no swagger
documentation will be generated for the tagged example groups. It works on all
kinds of example groups (responses, verbs, paths etc..).

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,12 +454,30 @@ end
 
 ### Running tests without documenting ###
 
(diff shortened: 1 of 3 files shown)


BookOfGreg:
This looks good, and tested and documented, thank you for this contribution! I'm going to take a day to think of the key 'document' is the right keyword to describe this before merging.

BookOfGreg:
I was considering if this word could be used for other things, other synonyms, or using the negative of this word but I think your choice is the best one :+1:

PR #263, opened 2019-12-16
Override Content Security Policy for Swagger UI Compat

### LINKED ISSUE #174: Content Security Policy for UI middleware
I am getting a problem where the UI is blank once I enable the CSP that is part of Rails 5.2. The reason is that the inline script tags which load the UI are blocked since I don't have `unsafe-inline` in my `script-src`.

Rails has the ability to let a controller override the CSP for a certain context but since the UI is served via middleware we don't have that option to override. I can add `unsafe-inline` to my global CSP and the problem is fixed but then my entire app allows inline scripts. :(

When looking at the Rails CSP module I noticed that [if the page already has a CSP header it won't add the global config](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/content_security_policy.rb#L21). I'm thinking this is the answer. Figure out what CSP header is appropriate for Swagger UI and have the RSwag middleware include that header.

I can do the leg work on making a PR for this but wanted to confirm this is the desired strategy and something you consider in-scope for this project.

eric-hemasystems:
@BookOfGreg - You mentioned over on #174 that you were open to a fix for this so I went ahead and put together a PR. This should provide that impl. When testing I realized that my `unsafe-eval` from #174 was not really needed. It was my Vue plugin giving me a false positive so the CSP for this PR is a bit tighter in terms of permissions.

This closes #174.

---

The Content Security Policy being served by Rails might be incompatible with what Swagger UI needs. This is true of the default CSP that Rails provides and even if it wasn't, a developer might configure their CSP to be incompatible with the Swagger UI.

The ensure the Swagger UI works it makes the most sense to have it provide it's own CSP. This will allow us to ensure we have all the permissions that UI needs while at the same time having no more than that.

The CSP I propose is:

```
default-src 'self';
img-src 'self' data: https://online.swagger.io;
```

The main sticking points are the `unsafe-inline` options for both styles and scripts. This is needed because [the `index.html` file](https://github.com/rswag/rswag/blob/master/rswag-ui/lib/rswag/ui/index.erb) has both inline styles and scripts. This is a permission most applications would be hesitant to enable on their app. Therefore if we ONLY enable it in this context we can allow the rest of the Rails app to be as locked down as the developer desires.

Since we are overriding the CSP we might as well do the most restrictive config that still works with the Swagger UI. For example the default CSP in Rails allows any font, image, style or script as long as it is hosted on a HTTPS server. It also allows data URLs for fonts and images.

For the Swagger UI we only need to allow data urls for images (the Swagger logo). The only 3rd party image is on https://online.swagger.io (for the validator image). Finally we need to allow fonts and styles to some Google sites for the Google fonts.

I suggest future PRs work on removing the inline styles and scripts from the `index.html` file. This will allow us to lock down the permissions further by removing the dangerous `unsafe-inline` permission. For now my main goal was compatibility with the current functionality. Since the `index.html` is based on an upstream system I wasn't sure if it was best to tackle that there. Also the upgrade to Swagger 3 probably plays into this as that file will likely change so perhaps we wait until Swagger 3 is working with RSwag before we work on locking down `index.html`.


BookOfGreg:
Recently merged PR #187 allows custom headers in the API section, I think a combination of these approaches would be best, providing a default CSP with the opportunity to overwrite it.

Will use this as a base for adding custom headers at a later time, or you can do so yourself.

Thank you for your contribution to the project so far.

eric-hemasystems:
Would that hook work for this? It is the delivery of the index page that needs the CSP which is a different middleware (api middleware vs ui middleware).

eric-hemasystems:
Since there was interest in this PR again on #174 I have updated my branch to merge with the latest in `master`. I gave it a whirl in the `test-app` bundled with the repo and it seems to work just fine. Looks like Swagger removed the validator image at the bottom so we can actually remove one entry from the CSP.

`unsafe-inline` for both styles and scripts are still required. Just like in my original PR if someone wants to tackle that (probably easiest way is via a nonce) it could be as part of a separate PR. My goal was just for rswag to be compatible with Rails when a CSP is enabled in Rails that is more restrictive than rswag needs.

jtannas (approved):
:tada:

RomainKoszyk:
Hi @jtannas,

Had the same issue with the current release `2.6.0`, do you know when you'll prepare a new release containing this [commit](https://github.com/rswag/rswag/commit/97c89d73561f64573435083b25227be5514f565e)?

Thanks!

jtannas:
@RomainKoszyk 

I've just created a git tag for the release and a discussion for it here: https://github.com/rswag/rswag/discussions/561

PR #274, opened 2020-01-30
Add ENV var to configure dry run

igas:
During integration with rswag, we noticed that we need to set dry run very early and suggestion is to put it in `application.rb`. We did not really want to add specs related code to production code. So we came up with this solution. What do you think?

### DIFF
--- a/rswag-specs/lib/rswag/specs/configuration.rb
+++ b/rswag-specs/lib/rswag/specs/configuration.rb
@@ -26,9 +26,11 @@ module Rswag
       end
 
       def swagger_dry_run
-        @swagger_dry_run ||= begin
-          @rspec_config.swagger_dry_run.nil? || @rspec_config.swagger_dry_run
+        return @swagger_dry_run if defined? @swagger_dry_run
+        if ENV.key?('SWAGGER_DRY_RUN')
+          @rspec_config.swagger_dry_run = ENV['SWAGGER_DRY_RUN'] == '1'
         end
+        @swagger_dry_run = @rspec_config.swagger_dry_run.nil? || @rspec_config.swagger_dry_run
       end
 
       def swagger_format
--- a/README.md
+++ b/README.md
@@ -452,14 +452,15 @@ after do |example|
   example.metadata[:response][:examples] = { 'application/json' => JSON.parse(response.body, symbolize_names: true) }
 end
 ```
-You need to disable --dry-run option for Rspec > 3
+You need to disable `--dry-run` option for Rspec > 3. You can do one of the following:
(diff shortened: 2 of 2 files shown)


rswag-specs/lib/rswag/specs/configuration.rb
```diff
@@ -26,9 +26,11 @@ def swagger_docs
       end
 
       def swagger_dry_run
-        @swagger_dry_run ||= begin
-          @rspec_config.swagger_dry_run.nil? || @rspec_config.swagger_dry_run
+        return @swagger_dry_run if defined? @swagger_dry_run
+        if ENV.key?('SWAGGER_DRY_RUN')
+          @rspec_config.swagger_dry_run = ENV['SWAGGER_DRY_RUN'] == '1'
```
akabiru, line 31: In order to maintain a familiar API, WDYT about keeping the env var value as a boolean?

```suggestion
          @rspec_config.swagger_dry_run = ENV['SWAGGER_DRY_RUN'] == true
```

akabiru (commented):
Thank you for this suggestion @igas I agree that having spec logic in `application.rb` seems a bit off the mark- _I've had to do the same as well_ 😄 .

This looks like a very good improvement- I would only request that you update [this section in the README](https://github.com/rswag/rswag#enable-generation-examples-from-responses) to reflect the change. cc @BookOfGreg

rswag-specs/lib/rswag/specs/configuration.rb
igas: Unfortunately, all ENV vars come as strings.

igas:
Hi @BookOfGreg, any chance you can take a look at this one?

BookOfGreg:
Code looks ok. I'd like to see a test related to the change. I'll probably merge either this or something similar after OA3 changes as that's the majority of the effort just now.

igas:
No worries, thanks a lot for work on OA3, we can't wait to test it out.

flyfy1:
Looking forward for sth like this to be merged, cuz the current master's README section about DryRun doesn't work.

The issue with current master branch is that, when doing `rails rswag`, the task param is generated first. However, the file containing the dry-run setting (for example, `environment/test.rb`) haven't been loaded at that time

BookOfGreg (approved):

gerardo-navarro:
May I ask what the status of this feature is? It seems to be approved and ready to be merged?

oblakeerickson:
@gerardo-navarro thanks for the bump. This feature is now in the 2.4.0 release.

PR #285, opened 2020-03-22
Update README.md

andymond:
Update path parameter example to use schema syntax - the old syntax was causing the react frontend to not load the parameter component


BookOfGreg:
Hey @andymond the old syntax should be valid swagger 2.0 syntax, are you seeing a version of swagger-ui that does not load that correctly? If you could point out a version of swagger-ui, or a swagger document that you could share I could validate this and perhaps check if it's not a bug there also.
Really appreciate the contribution, this might be quite interesting.

akofink:
OpenAPI 3 changed the parameter format to what is used in this PR: http://spec.openapis.org/oas/v3.0.3#parameter-object

rswag-specs needs to be updated to use this, though. Without `type: :type`, it calls `nil.to_sym` and fails: https://github.com/rswag/rswag/blob/master/rswag-specs/lib/rswag/specs/request_factory.rb#L121. This is just one example; I'm not sure if there are other issues.

BookOfGreg:
Nice try stale bot but I still want to look at this one (sooner or later)

PR #289, opened 2020-03-23
fix README.md: dry_run goes into test.rb, not application.rb

bjelline:
adding `RSpec.configure` to `application.rb` can lead to problems in production, where rspec may not be present.  The error message I got when using  `application.rb` and deploying to heroku was:

```
remote:  !     Could not detect rake tasks
remote:  !     ensure you can run `$ bundle exec rake -P` against your app
remote:  !     and using the production group of your Gemfile.
remote:  !     rake aborted!
remote:  !     NameError: uninitialized constant RSpec
...
remote:  !     /tmp/build_1c../config/application.rb:23:in `<top (required)>'
```

Using  `config/environments/test.rb` instead avoids these problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,7 +454,7 @@ end
 ```
 You need to disable --dry-run option for Rspec > 3
 
-Add to application.rb:
+Add to config/environments/test.rb:
 ```ruby
 RSpec.configure do |config|
   config.swagger_dry_run = false


BookOfGreg (approved):

dvanoni:
While I agree that `application.rb` does not seem to be the correct place to do this configuration—I too ran into the same problem in production—moving this into `config/environments/test.rb` makes the configuration no longer work correctly for me. When I run `rake rswag`, it always runs with the `--dry-run` option enabled and thus does not update the response examples in my Swagger document.

As far as I can tell, I believe the issue is that `config/environments/test.rb` is not loaded until *after* the rake task loads, resulting in the following cascade of effects:

1. Without the `config.swagger_dry_run = false` line executing, `swagger_dry_run` within the RSpec config will be `nil`.
1. The `swagger_dry_run` method in `Rswag::Specs::Configuration` will [return `true`](https://github.com/rswag/rswag/blob/7ceedab4cb7788ab6b86b96bd2d8f2419e11ebbb/rswag-specs/lib/rswag/specs/configuration.rb#L32).
1. When the rake task loads, the `true` value will [cause it to pass the `--dry-run` option to RSpec](https://github.com/rswag/rswag/blob/7ceedab4cb7788ab6b86b96bd2d8f2419e11ebbb/rswag-specs/lib/tasks/rswag-specs_tasks.rake#L15).

I've gotten around this issue by using the following code in `application.rb`:

```ruby
unless Rails.env.production?
  RSpec.configure do |config|
    config.swagger_dry_run = false
  end
end
```

However, I still don't think this is a great approach because `application.rb` just doesn't feel like the correct place to configure this.

I think `config/environments/test.rb ` is a better place to configure this, or even `swagger_helper.rb`, but I'm struggling to figure out an elegant way to load those before the rake task loads. I would love to hear your thoughts on this!

PR #291, opened 2020-03-24
request factory / add support of hash as query string params

p1-ra:
Hi, 

I've added support of hash to query parameters in the request factory.

With the following parameter definition: 

```
parameter name: :vFilter, :in => :query, :type => :hash, :required => false, schema: {
        '$ref': '#/definitions/....'
      }
```

With the following spec configuration: 
```
describe 'with risk filter' do
  (...)
  let(:vFilter) { {risks: ["FooBar"]} }
  run_test!
```

Will generate the following query parameters:  `"vFilter%5Brisks%5D%5B%5D=FooBar"`

Sorry, but I do not have the time to write new tests.
Best regards.

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -75,19 +75,27 @@ module Rswag
 
       def build_query_string_part(param, value)
         name = param[:name]
-        return "#{name}=#{value.to_s}" unless param[:type].to_sym == :array
 
-        case param[:collectionFormat]
-        when :ssv
-          "#{name}=#{value.join(' ')}"
-        when :tsv
-          "#{name}=#{value.join('\t')}"
-        when :pipes
-          "#{name}=#{value.join('|')}"
-        when :multi
-          value.map { |v| "#{name}=#{v}" }.join('&')
+        case param[:type].to_sym 
+        when :array
+          case param[:collectionFormat].to_sym 
+          when :ssv
+            "#{name}=#{value.join(' ')}"
+          when :tsv
+            "#{name}=#{value.join('\t')}"
+          when :pipes
+            "#{name}=#{value.join('|')}"
+          when :multi
+            value.map { |v| "#{name}=#{v}" }.join('&')
+          else
+            "#{name}=#{value.join(',')}" # csv is default
+          end
+        when :hash
+          query_hash = {}
+          query_hash[name] = value
+          query_hash.to_query
         else
-          "#{name}=#{value.join(',')}" # csv is default
+          "#{name}=#{value.to_s}"
         end
       end
 


BookOfGreg:
Thanks for the contribution but I don't think this is what you're after.

You can already achieve this effect using `type: object`, and providing the schema for the json structure that is provided.
Type also refers to JSON Schema type which is I believe limited to JSON types.

See: https://swagger.io/specification/#parameterObject

And: https://swagger.io/specification/#schemaObject

If there's some way we can make the usage of this clearer then I'd accept a PR to update the readme to make your use case clearer or easier.

p1-ra:
> Thanks for the contribution but I don't think this is what you're after.
> 
> You can already achieve this effect using `type: object`, and providing the schema for the json structure that is provided.
> Type also refers to JSON Schema type which is I believe limited to JSON types.
> 
> See: https://swagger.io/specification/#parameterObject
> 
> And: https://swagger.io/specification/#schemaObject
> 
> If there's some way we can make the usage of this clearer then I'd accept a PR to update the readme to make your use case clearer or easier.

Hi,

Thanks for the hint,  may I ask a little example on how to use it with rswag-spec ?

By using the `:object` type:  
```
parameter name: :vFilter, :in => :query, :type => :object, :required => false, schema: {
        type: 'object',
        properties: {
          risk_levels: {
            type: 'array',
            items: {
              type: 'string'
            }
          }
        }
      }
```

I've noticed that:

1. Schema structure seem to not be use. If I set a value that violate its schema definition, no error is triggered about it.

2.  When using `:type => :object`  it generate an invalid URI:
```
# Object used as parameter
let(:vFilter) { {risk_levels: ['FooBar']}

# Triggered error
Failures:

  1) (...) risk_level filter returns a 200 response
     Failure/Error: raise InvalidURIError, "bad URI(is not URI?): #{uri}"

     URI::InvalidURIError:
       bad URI(is not URI?): http://www.example.com/....?vFilter={:risk_levels=%3E[%22FooBar%22]}
     # /home/p1sec/.gem/ruby/2.5.5/gems/rack-2.2.2/lib/rack/mock.rb:94:in `parse_uri_rfc2396'
     # /home/p1sec/.gem/ruby/2.5.5/gems/rack-2.2.2/lib/rack/mock.rb:106:in `env_for'
     # /home/p1sec/.gem/ruby/2.5.5/gems/rack-test-1.1.0/lib/rack/test.rb:262:in `env_for'
     # /home/p1sec/.gem/ruby/2.5.5/gems/rack-test-1.1.0/lib/rack/test.rb:118:in `request'
     # /home/p1sec/.gem/ruby/2.5.5/gems/rswag-specs-2.2.0/lib/rswag/specs/example_helpers.rb:19:in `submit_request'
     # /home/p1sec/.gem/ruby/2.5.5/gems/rswag-specs-2.2.0/lib/rswag/specs/example_group_helpers.rb:87:in `block in run_test!'
```

Regarding the code of the `RequestFactory` class, it seem that their is actually no custom handling for the `:object` type during the build of the query parameters https://github.com/rswag/rswag/blob/b70e3a0794c056f93d5b01b7ccbafae40c1fe2c9/rswag-specs/lib/rswag/specs/request_factory.rb#L76-L92

Thanks in advance

paulomcnally:
Any solution for this case?

paulomcnally:
ping 
@BookOfGreg

PR #298, opened 2020-04-15
Fix route_parser each_with_object call

julianrubisch:
When I tried to generate a new spec, I noticed the `each_with_object` call in `route_parser.rb` was mixed up.

### DIFF
--- a/rswag-specs/lib/rswag/route_parser.rb
+++ b/rswag-specs/lib/rswag/route_parser.rb
@@ -11,7 +11,7 @@ module Rswag
     def routes
       ::Rails.application.routes.routes.select do |route|
         route.defaults[:controller] == controller
-      end.each_with_object({}) do |tree, route|
+      end.each_with_object({}) do |route, tree|
         path = path_from(route)
         verb = verb_from(route)
         tree[path] ||= { params: params_from(route), actions: {} }


BookOfGreg:
Is it possible for you to add a test for this to illustrate the problem?
Your code looks correct at a glance but tests ensure I don't break it again accidentally in the future :)

julianrubisch:
Sure, that’s a good idea. I just fixed it on the fly because I needed it to work on a project...

mathieujobin:
I just got bitten by this... I wondered if it was a change in ruby 2.7 ?!?

I am surprise there is no test covering the rake tasks?

there should be a release with this fix.

`rails generate rspec:swagger Api::V1::FooController` is broken as of 2.3.1

thank you

mathieujobin:
For anyone needing help to use the fork

```ruby
gem 'rswag-api'
gem 'rswag-ui'

group :development, :test do
  gem 'rswag-specs', github: 'julianrubisch/rswag', branch: 'fix-route-parser'
end
```

julianrubisch:
Haven’t managed to write the test yet 😕 as it would require lots of stubbing rails internals

BookOfGreg:
@mathieujobin 
> I am surprise there is no test covering the rake tasks?

There are tests covering it.
If you would be so kind as to add a test for this section yourself we may be able to get this merged.
A unit test covering the route_parser.rb specifically would be enough.

Thank you.

mathieujobin:
@BookOfGreg I might in the future, but at this point, I am just trying out this gem for the first time. 
I would first want to be able to use it before contributing.

BookOfGreg:
Merged #324 which contained tests.
Thank you for your effort here.

PR #301, opened 2020-04-18
Allow to use different securities per response block

sobrinho:
Imagine that you want to test a specific endpoint where the you must not
send the credentials, like a sign in or recovery password endpoint where
you must not be logged in.

The previous implementation was leaking the security definition to
unrelated responses due to the RSpec metadata design.

This is still true for other definitions but the most important one to
not leak is just the security.


sobrinho:
not stale

PR #302, opened 2020-04-18
[Fix] move examples inside mime_type

pedrocarmona:
**Fixes**
Currently for openapi: 3.0.1, the examples are not being rendered because the 
the key `examples` was moved into inside mime_type.

This PR also adds example method.
This is only available in openapi: 3.0.1.

### DIFF
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -126,16 +126,21 @@ module Rswag
         target_node = metadata[:response]
         upgrade_content!(mime_list, target_node)
         metadata[:response].delete(:schema)
+        metadata[:response].delete(:examples)
       end
 
       def upgrade_content!(mime_list, target_node)
         target_node.merge!(content: {})
         schema = target_node[:schema]
+        examples = target_node[:examples]
         return if mime_list.empty? || schema.nil?
 
         mime_list.each do |mime_type|
           # TODO upgrade to have content-type specific schema
-          target_node[:content][mime_type] = { schema: schema }
+          target_node[:content][mime_type] = {
+            schema: schema,
+            examples: examples,
+          }
         end
       end
 
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -181,7 +181,7 @@ module Rswag
       def upgrade_servers!(swagger_doc)
         return unless swagger_doc[:servers].nil? && swagger_doc.key?(:schemes)
 
-        #ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: schemes, host, and basePath are replaced in OpenAPI3! Rename to array of servers[{url}] (in swagger_helper.rb)')
+        ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: schemes, host, and basePath are replaced in OpenAPI3! Rename to array of servers[{url}] (in swagger_helper.rb)')
 
         swagger_doc[:servers] = { urls: [] }
         swagger_doc[:schemes].each do |scheme|
@@ -201,14 +201,14 @@ module Rswag
         schemes.each do |name, v|
           next unless v.key?(:flow)
 
-          #ActiveSupport::Deprecation.warn("Rswag::Specs: WARNING: securityDefinitions flow is replaced in OpenAPI3! Rename to components/securitySchemes/#{name}/flows[] (in swagger_helper.rb)")
+          ActiveSupport::Deprecation.warn("Rswag::Specs: WARNING: securityDefinitions flow is replaced in OpenAPI3! Rename to components/securitySchemes/#{name}/flows[] (in swagger_helper.rb)")
           flow = swagger_doc[:components][:securitySchemes][name].delete(:flow).to_s
           if flow == 'accessCode'
-            #ActiveSupport::Deprecation.warn("Rswag::Specs: WARNING: securityDefinitions accessCode is replaced in OpenAPI3! Rename to clientCredentials (in swagger_helper.rb)")
+            ActiveSupport::Deprecation.warn("Rswag::Specs: WARNING: securityDefinitions accessCode is replaced in OpenAPI3! Rename to clientCredentials (in swagger_helper.rb)")
             flow = 'authorizationCode'
           end
           if flow == 'application'
-            #ActiveSupport::Deprecation.warn("Rswag::Specs: WARNING: securityDefinitions application is replaced in OpenAPI3! Rename to authorizationCode (in swagger_helper.rb)")
+            ActiveSupport::Deprecation.warn("Rswag::Specs: WARNING: securityDefinitions application is replaced in OpenAPI3! Rename to authorizationCode (in swagger_helper.rb)")
             flow = 'clientCredentials'
           end
           flow_elements = swagger_doc[:components][:securitySchemes][name].except(:type).each_with_object({}) do |(k, _v), a|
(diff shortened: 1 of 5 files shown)


rswag-specs/spec/rswag/specs/swagger_formatter_spec.rb
```diff
@@ -167,6 +167,85 @@ module Specs
...
+                  schema: { '$ref' => '#/definitions/blog' },
+                  examples: {
+                    one: {
+                      value: {
+                        'data' => {
+                          'id' => '1',
+                          'type' => 'record'
+                        }
+                      }
+                    },
```
pedrocarmona, line 260: allows to use in the integration test:
``` 
...
examples(
  one: {
    value: {
      'data' => {
        'id' => '1',
        'type' => 'record'
      }
    }
  }
)
...
```

chunlea:
From https://swagger.io/docs/specification/adding-examples/#request-response-bodies I think we should also support `example`, not only `examples`.

pedrocarmona:
> From https://swagger.io/docs/specification/adding-examples/#request-response-bodies I think we should also support `example`, not only `examples`.

Shall I add new method `example` to `/rswag-specs/lib/rswag/specs/example_group_helpers.rb`?

chewi:
Seems we've done roughly the same thing at roughly the same time. See #305.

pedrocarmona:
this PR works for both versions, but is not fully compliant, as the user needs to update the code.
Its required to remove the mime type present in the `examples(` usage.

Alternatively, we could keep the same api, but I think its legacy.
In my opinion, the mime type should be registered automatically from the request performed, for both the schema and the responses examples.`

Also we are not registering the request examples (available in openapi: 3.0.1)

chewi:
Bear in mind that there might not always be a real request/respond to base the MIME types on. Ideally there should be but I haven't had a chance to stub out all the many parts of my application.

Regarding the request examples, I submitted #304 to help with that but you still need some code to link it up. This is roughly what open-api-rswag did.

```ruby
  after do |example|
    if response
      if response.code.to_s =~ /^2\d{2}$/ &&
         (body_param = example.metadata[:operation][:parameters].find { |p| p[:in] == :body })
        body_param[:examples] = {
          body_param[:name] => { value: send(body_param[:name]) }
        }
      end
    end
  end
```

mia-n:
If this is the 3.0 Response Example support PR that y'all end up going with, could you still also add support for automatically generated response examples? The README still references this here:

https://github.com/rswag/rswag/blob/master/README.md#enable-auto-generation-examples-from-responses

There just needs to be a soft check in swagger_formatter to see if content already exists on the response before assigning over it. I created a solution (which would conflict with this solution) that would allow for this in #307. I'm happy to close #307 and abandon those changes if we could add just that part allowing auto-generated response examples in here.

rswag-specs/lib/rswag/specs/swagger_formatter.rb
```diff
@@ -124,19 +124,44 @@ def upgrade_response_produces!(swagger_doc, metadata)
...
-      def upgrade_content!(mime_list, target_node)
-        target_node.merge!(content: {})
+      def upgrade_schema!(mime_list, target_node)
         schema = target_node[:schema]
         return if mime_list.empty? || schema.nil?
 
+        set_mime_list_contents!(target_node, mime_list, :schema, schema)
+      end
+
+      def upgrade_examples!(mime_list, target_node)
```
mia-n, line 142: @pedrocarmona I pulled your code and wired it to my own project - as it is, schemas generated from your fork of rswag with examples added in this way will fail openapi_cli validation, because you're dumping the contents of the old-style [metadata][examples] into [metadata][content][mime_type][examples] without handling the contents.

If the rswag-consuming application implements the suggested code for automatic response examples (which is:
```
after do |example|
end
```
per https://github.com/rswag/rswag#enable-auto-generation-examples-from-responses)

then their schema has [metadata][examples][mime-type]. Your code then dumps this into [metadata][content][mimetype][examples][mime-type][actual example].

You end up with a result that looks like:
```
responses:
        '200':
          description: returns data
          content:
            application/json:
              examples:
                application/json:
                  some_actual_response_content: "asdasdasda"
```

which, on running the openapi validator (I'm running it through docker as per https://openapi-generator.tech/docs/installation/#docker with `docker run --rm -v [APP_ROOT]/local openapitools/openapi-generator-cli:v4.2.3 validate --input-spec /local/[MY_SCHEMA_DIR]` if you want 100% reproducibility)

you get reported schema errors like:
```
Validating spec (/local/openapi/v1/openapi.yaml)
Errors:
        -attribute paths.'/spaces'(post).responses.200.content.examples.application/json.[CONTENTS] is unexpected
```

Fixing this is going to be tricky, and I'm not sure what the right path forward is. Setting the mime list contents needs to be less naive about the contents if it's important to keep 100% backwards compatibility with the way the README says to add examples.

mia-n: A solution that I implemented in my own fork (based off your fork now):
changing swagger_formatter.rb method `upgrade_examples!` to:

```
def upgrade_examples!(mime_list, target_node)
        example = target_node[:example]
        examples = target_node[:examples]

        if example
          set_mime_list_contents!(target_node, mime_list, :example, example)
        elsif examples
          #check if examples contains mime_types from swagger2.0 response example style
          non_mime_type_examples = {}
          examples.each do |key, value|
            if mime_list.include? key
              set_mime_type_content!(target_node, key, :examples, value)
            else
              non_mime_type_examples[key] = value
            end
          end
          set_mime_list_contents!(target_node, mime_list, :examples, non_mime_type_examples)
        end
      end
```

I don't know if this solution will work universally, but it should:
* Allow the consumer to use the current README suggested method of adding autogenerated response examples.
* Override those autogenerated examples with integration-spec-level-defined custom example responses
* Do so in a way that is compliant with the openapi-generator-cli validator.

Would it be valuable for me to open a PR to your fork with this, or...open a PR here, or something else?

pedrocarmona: yes, I think I will need to add a bit more tests on top of your PR.

mia-n: Opened a PR on your fork instead of here? I figured that would be faster + more helpful, if you'd rather I open it here above yours instead lmk

pedrocarmona: merged

pernambucano:
As this hasn't been applied yet, I'd like to point out that there is another pr for the same thing:  #330.

chewi:
So that makes three of them now? This is getting silly.

pernambucano:
I guess there's another one: #288 .  Although, it is more concerned about the schema. But I think it has a common theme of "things that should be inside the content but aren't yet".
PS: I'm just trying to minimize the work that will need to be done when we finish this PR (closing all the other too).

phlegx:
**Summary of similar open PR requests:**

- #345
- #344 (Fix by modifying `upgrade_content!` method in Swagger Formatter)
- #330 (Fix by modifying `upgrade_content!` method in Swagger Formatter)
- #305 (This is more a feature request than a fix)

Solved in version 2.3.1
- #288

There are different procedures, but which is the right one?

sergioisidoro:
Any update on this issue? I think I'm also affected, and would like to help if possible

sergioisidoro:
@pedrocarmona @chewi is it ok to try and focus efforts on https://github.com/rswag/rswag/pull/305 , and close this one? Or the other way arround?

chewi:
I'm almost past caring at this point. We do still use my fork, but it's not used in an important area, so I can't really justify spending further time on this.

pedrocarmona:
ok go ahead :)

PR #304, opened 2020-04-20
Support more than just schema in requestBody and handle formData params better

chewi:
Both body and formData parameters can have descriptions, examples, and a required flag.

There can only be one body parameter but there can be multiple formData parameters, each with their own names and schemas.

This causes one of the existing tests to fail but I have doubts about whether the test case is valid OpenAPI. I can't see an example quite like this in the OpenAPI spec. What do you think?

```yaml
content:
  multipart/form-data:
    schema:
      type: file
```


phlegx:
This PR works for me! I use request body examples and they are generated correctly in swagger file!

PR #305, opened 2020-04-20
Allow multiple response examples to be defined per status code

chewi:
For example:

```ruby
  after do |example|
    example.example_group.example_value JSON.parse(response.body, symbolize_names: true) if response
  end

  response 422, 'invalid request' do
    example_context 'name is missing'
      let(:password) { 'secret' }
      run_test!
    end

    example_context 'password is missing' do
      let(:name) { 'bobby' }
      run_test!
    end

    example_context 'bad content type' do
      example_value(
        error: 'bad content type'
      )
    end
  end
```

Annoyingly you still need a dummy RSpec example to include a static example value on its own but that's a limitation of RSwag's design.

I really really want to be able to have multiple media types per response here and the lack of support for that is killing me. I tried to see whether there was a straightforward way to add it but didn't get anywhere. It's a shame as this worked in the openapi-rswag fork. Trying to shoehorn V3 support into V2 syntax doesn't seem like the best way to go.

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'active_support/core_ext/string/inflections'
+
 module Rswag
   module Specs
     module ExampleGroupHelpers
@@ -52,8 +54,19 @@ module Rswag
       end
 
       def response(code, description, metadata = {}, &block)
-        metadata[:response] = { code: code, description: description }
+        metadata[:response] = { code: code, description: description, examples: {} }
         context(description, metadata, &block)
+      end
+
+      def example_context(description, metadata = {}, &block)
+        context(description, metadata.merge(skip_formatter: true), &block)
+      end
+
+      def example_value(value)
+        metadata[:response][:examples][metadata[:description].parameterize separator: '_'] = {
+          summary: metadata[:description],
+          value: value
+        }
       end
 
       def schema(value)
--- a/README.md
+++ b/README.md
@@ -597,7 +597,7 @@ To enable examples generation from responses add callback above run_test! like:
 
 ```
 after do |example|
-  example.metadata[:response][:examples] = { 'application/json' => JSON.parse(response.body, symbolize_names: true) }
+  example.example_group.example_value JSON.parse(response.body, symbolize_names: true)
 end
 ```
 
--- a/rswag-specs/lib/generators/rspec/templates/spec.rb
+++ b/rswag-specs/lib/generators/rspec/templates/spec.rb
@@ -19,7 +19,7 @@ RSpec.describe '<%= controller_path %>', type: :request do
 <%      end -%>
 
         after do |example|
-          example.metadata[:response][:examples] = { 'application/json' => JSON.parse(response.body, symbolize_names: true) }
+          example.example_group.example_value JSON.parse(response.body, symbolize_names: true)
         end
         run_test!
       end
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/deep_merge'
+require 'active_support/core_ext/hash/slice'
+require 'active_support/core_ext/object/blank'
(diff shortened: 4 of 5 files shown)


jmajonis:
@chewi Do you have advice how to try your feature branch in a local Rails project?

I've tried bundling the following from the Gemfile:

```
gem 'rswag', git: 'https://github.com/chewi/rswag.git', branch: 'response-examples'
```
and 

```
git 'https://github.com/chewi/rswag.git', branch: 'response-examples' do
  gem 'rswag-api'
  gem 'rswag-specs'
  gem 'rswag-ui'
end
```

But after building a new Docker image and restarting I get errors related to `swagger-ui`

```
Started GET "/api-docs/index.html" for 192.168.112.1 at 2020-05-12 17:19:39 +0000
ActionController::RoutingError (No route matches [GET] "/api-docs/swagger-ui.css"):
ActionController::RoutingError (No route matches [GET] "/api-docs/swagger-ui-standalone-preset.js"):
ActionController::RoutingError (No route matches [GET] "/api-docs/swagger-ui-bundle.js"):
ActionController::RoutingError (No route matches [GET] "/api-docs/swagger-ui-bundle.js"):
ActionController::RoutingError (No route matches [GET] "/api-docs/swagger-ui-standalone-preset.js"):
```

I'm sure it's not your feature branch. I'm misunderstanding something about how to bundle from a Git source.

chewi:
This is an issue with rspec-ui, which I haven't changed and do not use. You cannot use it directly from git because `npm install` needs to be called to put the swagger-ui-dist files in place.

rswag-specs/lib/rswag/specs/example_group_helpers.rb
```diff
@@ -52,10 +54,21 @@ def parameter(attributes)
...
+        metadata[:response] = { code: code, description: description, examples: {} }
         context(description, metadata, &block)
       end
 
+      def example_context(description, metadata = {}, &block)
+        context(description, metadata.merge(skip_formatter: true), &block)
+      end
+
+      def example_value(value)
+        metadata[:response][:examples][metadata[:description].parameterize separator: '_'] = {
```
ai-oleynikov, line 66: In `ActiveSupport` < 5 this method had different signature: `parameterize(sep = '-')`
And since we [support](https://github.com/BookOfGreg/rswag/blob/10bb73214882c019d91e0e48d01e6aacaa03ceb1/rswag-specs/rswag-specs.gemspec#L16) versions >= 3.1, I suggest to drop the argument and stick with the default separator.

rswag-specs/lib/rswag/specs/example_group_helpers.rb
```diff
@@ -52,10 +54,21 @@ def parameter(attributes)
...
-        metadata[:response] = { code: code, description: description }
+        metadata[:response] = { code: code, description: description, examples: {} }
         context(description, metadata, &block)
       end
 
+      def example_context(description, metadata = {}, &block)
+        context(description, metadata.merge(skip_formatter: true), &block)
+      end
+
+      def example_value(value)
```
ai-oleynikov, line 65: Wouldn't it be better to add `metadata[:response][:examples] ||= {}` here instead of [changing](https://github.com/rswag/rswag/pull/305/files#diff-7c43e7d1fb91b5c86d12f8c431b704ea535c00e16694677fda4ecb67b9277fc7R57)
 `#response`?

This would be consistent with [parameters](https://github.com/rswag/rswag/blob/8c4be0f247b0831c6cbcc5e8dc6f7ede684225e8/rswag-specs/lib/rswag/specs/example_group_helpers.rb#L48) and [headers](https://github.com/rswag/rswag/blob/8c4be0f247b0831c6cbcc5e8dc6f7ede684225e8/rswag-specs/lib/rswag/specs/example_group_helpers.rb#L77). Also, this way you wouldn't have to [change existing tests](https://github.com/rswag/rswag/pull/305/files#diff-940d36a104df7df89754f468e7b258ea77564d5abf68151eb0122b66f78297b9R114). And resulting Swagger documents won't have these empty `examples: {}` for every response without examples.

ai-oleynikov (commented):
Hey, I'm not a maintainer, but I tried your PR and I have a couple of comments.

sergioisidoro:
Another duplicate: https://github.com/rswag/rswag/pull/345

sergioisidoro:
Should be resolved by: https://github.com/rswag/rswag/pull/501

PR #312, opened 2020-05-03
Update README.md

levidavidmurray:
Update **Referenced Parameters and Schema Definitions** example to use referenced parameters.

The existing README isn't exactly clear on how to use references for parameters such as requestBody for a post request. Currently, you can really only infer it from the `oneOf` example, or dive into the test-app examples.


oblakeerickson (approved):
This looks good to me :+1: Should be safe to merge

BookOfGreg:
Brilliant, thanks! :)

PR #322, opened 2020-05-22
Add a license

chrisf:
rswag doesn't currently specify a license, so it would be nice to have one

https://choosealicense.com/licenses/mit/

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 rswag
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


flyfy1:
Hi @chrisf, the license is in each of the sub-folder for this library

chrisf:
Don’t know how I missed that! If that’s the case, maybe we can add a mention in the README in the root folder so it’s easier to discover. 

Since it looks like they’re all the same license, would it be possible to add one to the root folder as well so GitHub detects it?

flyfy1:
haha I think it's a good idea but I'm not the repo's owner..

BookOfGreg:
I got you fam 👍

Issue #323, opened 2020-05-24
undefined method `path' for {}:Hash (NoMethodError)


graceful-potato:
When I try to create an integration spec using a generator for my controller, I get an error `gems/rswag-specs-2.3.1/lib/rswag/route_parser.rb:26:in `path_from': undefined method `path' for {}:Hash (NoMethodError)`

Fix - #324

deniciocode:
Is this issue fixed? I have the same error.

z0rk1i:
Works for **rails generate rspec:swagger API::OrdersController**
Fails with same error for **rails generate rspec:swagger API::V1::OrdersController**

deniciocode:
@z0rk1i do you have a route for your `V1`? Do you have the same issue as mentioned in the issue for `V1`?

graceful-potato:
@z0rk1i 
I think this means that you don't have API::OrdersController in your application or you haven't set the routes for this controller. Because when you pass a real controller to the generator, and you have routes that use that controller, an exception will be raised

And I am pretty sure that generated spec for your API::OrdersController is completely empty even if you have actions in the controller.

z0rk1i:
Yes, it was just my mistake **rails generate rspec:swagger Api::OrdersController** works fine

z0rk1i:
After more detailed testing I see what it works with non existing controllers and generates empty spec.
If controller with route exist - it throws mentioned error.

jrmhaig:
I'm also seeing this. I have controllers `Reservations::V1Controller` and `Reservations::V2Controller` with routes defined as:

```ruby
  scope :api do
    namespace :reservations do
      resources :v1, param: :registration
      resources :v2, param: :registration
    end
  end
```

I get the error when I try

```bash
bundle exec rails g rspec:swagger Reservations::V1Controller
```

and I'm wondering if the `scope :api` is causing a problem, as this is adding `api/` to the path, which doesn't exist in the controller name.

**Update:** I think that this would be fixed by #352

BookOfGreg:
Fixed by #324

mikeheft:
I still get this error when trying to run `rails g rspec:swagger Api::V1::UsersController`. Here is the definition of my routes 
```ruby
api_version(module: 'Api::V1', header: { name: 'Accept', value: 'application/vnd.dispatchhealth.com; version=1' }, defaults: { format: :json }) do
    scope path: '/api', constraints: { format: /(json)/ } do
```

brendonrapp:
[...]

PR #324, opened 2020-05-24
Fix routes method in RoutePareser

graceful-potato:
Fix for #323

### DIFF
--- a/rswag-specs/lib/rswag/route_parser.rb
+++ b/rswag-specs/lib/rswag/route_parser.rb
@@ -11,7 +11,7 @@ module Rswag
     def routes
       ::Rails.application.routes.routes.select do |route|
         route.defaults[:controller] == controller
-      end.each_with_object({}) do |tree, route|
+      end.each_with_object({}) do |route, tree|
         path = path_from(route)
         verb = verb_from(route)
         tree[path] ||= { params: params_from(route), actions: {} }
--- a/rswag-specs/spec/rswag/route_parser_spec.rb
+++ b/rswag-specs/spec/rswag/route_parser_spec.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+RSpec.describe Rswag::RouteParser do
+  describe "#routes" do
+    let(:controller) { "api/v1/posts" }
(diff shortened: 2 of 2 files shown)


flyfy1:
Thx, I applied your fix locally and it helps resolving exactly the same issue that I'm facing.

Interestingly we could encounter such kind of issue here..

graceful-potato:
According to the git blame they changed `reduce` to `each_with_object` because of rubocop, and forgot that the block's signature is different for these methods.
But the fact that this bug has been here for two months is a bit weird.

VoltanBro:
Thx u bro!

flyfy1:
Hi @BookOfGreg could u take a look into this one?

deniciocode:
Hey,
I also applied these changes locally and I am able to use the generator now.

andrepcg:
Thanks, it fixes it!

graceful-potato:
@BookOfGreg

ronail:
works like a charm! When can we have this merged?

maycowmeira (approved):

dhivyada:
Please merge this to the base repo soon and release

drselump14:
When this bug fixed will be released?

BookOfGreg:
> 
> 
> According to the git blame they changed `reduce` to `each_with_object` because of rubocop, and forgot that the block's signature is different for these methods.
> But the fact that this bug has been here for two months is a bit weird.

Yeah that's really embarrassing. Sorry everyone!

LE-HU:
Guys I have an impression this problem persists in current 2.3.1 release. I've just started a new project and had to do the fix locally in order for generator to work.

Okay i just realized 2.3.1 was released in April and the fix is from May, shouldn't this fix be published so it's fixed automatically? Idk if what i'm saying makes sense as I'm pretty new around here, but as far as I understand everyone installing rswag currently have to do the fix manually.

jimbali:
@LE-HU Agreed. @BookOfGreg how do we go about getting this released please?

kamalpanhwar:
This issue still in latest version not seem not merged yet.

PR #330, opened 2020-06-04
Example responses handling

### LINKED ISSUE #325: OAS 3 Document has examples in the wrong path
According to the README:

> **Enable auto generation examples from responses**
> To enable examples generation from responses add callback above run_test! like:

```
after do |example|
end
```

However this results in the examples being directly on the `response` object, whereas it should be on the media type object of the response's `content` property.

```
Structural error at paths./api/v1/accounts.get.responses.200
should NOT have additional properties
additionalProperty: examples
Jump to line 158
```

Any suggestions for a way to work around this?

olegykz:
Seems like the current version of gem generates structure irrelevant to the 3.0 API. As a result, any response examples (dynamic or static) are not being shown on UI.

PR resolves https://github.com/rswag/rswag/issues/325

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,8 +363,8 @@ you should use the folowing syntax, making sure there are no whitespaces at the 
 
 ### Specifying/Testing API Security ###
 
-Swagger allows for the specification of different security schemes and their applicability to operations in an API. 
-To leverage this in rswag, you define the schemes globally in _swagger_helper.rb_ and then use the "security" attribute at the operation level to specify which schemes, if any, are applicable to that operation. 
+Swagger allows for the specification of different security schemes and their applicability to operations in an API.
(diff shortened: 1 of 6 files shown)


mediafinger:
We are first time users of rswag and have been wondering why the example generation (for `openapi: "3.0.1"`) does not work. We thought we had some misconfiguration on our side. Thanks for this PR, I hope and can be merged soon, as it seems to fix an issue several other people have.

Bandes:
I've been struggling with this as well, and I'm really hoping that this or a similar fix can be released soon. I really thought I was doing something wrong...

jamie (approved):
Of the PRs that fix response examples, this is the tightest. Merged to my WIP branch.

PR #332, opened 2020-06-05
Ensure `nullable` works for allOf, anyOf etc.

d11wtq:
Previously the `nullable` check was only being applied to the `type`
attribute validation, but schemas using `anyOf`, `allOf`, `oneOf` etc
were unable to use `nullable` because they would never be passed
directly to the field validator. This is crucial when using `$ref`
combined with `nullable`, because these cases require the use of
`allOf`.

The solution here is to simplify the validation extension at the root
and not to attach the `nullable` handling to any particular attribute.

Before this change, the following would fail validation if `foo` was
`null`:

```
{
  "type": "object",
  "properties": {
    "foo": {
      "nullable": true,
      "allOf": [
        {
          "$ref": "#/components/schemas/Foo"
        }
      ]
    }
  }
}
```

With this change the `nullable` check is handled at the right spot
before the `allOf` is evaluated.

### DIFF
--- a/rswag-specs/lib/rswag/specs/extended_schema.rb
+++ b/rswag-specs/lib/rswag/specs/extended_schema.rb
@@ -7,14 +7,11 @@ module Rswag
     class ExtendedSchema < JSON::Schema::Draft4
       def initialize
         super
-        @attributes['type'] = ExtendedTypeAttribute
         @uri = URI.parse('http://tempuri.org/rswag/specs/extended_schema')
         @names = ['http://tempuri.org/rswag/specs/extended_schema']
       end
-    end
 
-    class ExtendedTypeAttribute < JSON::Schema::TypeV4Attribute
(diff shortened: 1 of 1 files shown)


d11wtq:
Couldn't see any existing specs for the extension behaviour, but none of the current specs fail with this change.

brunocalmels:
Will this be merged for 2.5.0? Link says it's due for 2.4.0 but it actually points to 2.5.0 milestone. It'd be very helpful to have it released. Cheers!

erickreutz:
@jamie can we get this merged? nullable + $refs are broken

sergioisidoro:
Is it possible to add a test for the broken case, to avoid regressions?

sergioisidoro:
This was fixed by: https://github.com/rswag/rswag/pull/515 
Thanks @d11wtq 👍🏽

PR #345, opened 2020-07-15
Convert examples to match their expected behaviour in OpenAPI 3 specification

ghost:
At present, the `examples` generated by this gem are not compliant with the OpenAPI 3 specification and are hence ignored by tools that follow the spec (like Swagger Editor and Swagger UI). As seen [here](https://swagger.io/specification/), examples are expected to be at the media type level, similar to schema, and take an object with a `value` field.

This PR alters the `upgrade_content!` method of the swagger formatter so that it also upgrades the examlpes generated by the test, similar to how schema is handled.

Thank you, and let me know if you have any feedback.


smridge:
Tested and works like a charm!

eaylward8:
Just ran into the issue fixed by this PR, any idea when it might be merged?

sergioisidoro:
Should be resolved by: https://github.com/rswag/rswag/pull/501

PR #347, opened 2020-07-21
Fixes #346 - Enable Empty Response Body

### LINKED ISSUE #346: Describing Empty Response Body
Currently, describing an empty response body is not feasible. As mentioned in [documentation](https://swagger.io/docs/specification/describing-responses/#empty) `content` should be left out to enable responses without content.

I found the default insertion of an empty `content` block at [formatter](
https://github.com/rswag/rswag/blob/42fdf6d482b2e5d4aee4f461abf697b2aa9c3894/rswag-specs/lib/rswag/specs/swagger_formatter.rb#L132). To my mind, that hash should be initialized after the return statement in line 135.

psmandzich:


rswag-specs/lib/rswag/specs/swagger_formatter.rb
```diff
@@ -129,9 +129,9 @@ def upgrade_response_produces!(swagger_doc, metadata)
       end
 
       def upgrade_content!(mime_list, target_node)
-        target_node.merge!(content: {})
         schema = target_node[:schema]
         return if mime_list.empty? || schema.nil?
+        target_node.merge!(content: {})
```
olegykz, line 134: "content" key overriding causes https://github.com/rswag/rswag/issues/325

jamie (approved):
This fix is workable, added to my WIP pr having resolved the merge conflict with #330 (which @olegykz's comment here points to)

PR #349, opened 2020-07-22
Fix typo in test-app/app/controllers/blogs_controller.rb

MikeSmithEU:
Fix typo


oblakeerickson (approved):
Looks safe to merge in :+1: Thanks!

BookOfGreg:
Hmm, that piece must never be run for this typo to not cause problems.
Will merge but might add an extra test later.

PR #352, opened 2020-08-02
Fix route parser

drselump14:
It seems that  the order of parameter is wrong.

The documentation of `each_with_object` method

https://apidock.com/rails/Enumerable/each_with_object


jrmhaig:
I was literally just about to fork and create exactly the same PR. I thought I had better check what was already open first, though.

BookOfGreg:
Dup of #324
Sorry it took long enough for me to merge that it caused a dup.

drselump14:
Never mind @BookOfGreg , thanks for merging the fix PR

PR #358, opened 2020-09-02
Allow for '#/component/responses/' $ref in specs

tyler-wel:
This pull request allows generic shared response schemas to be referenced in specs, for example, referencing shared error responses.

This is inline with the [OpenAPI V3 Specifications](https://swagger.io/docs/specification/components/)

Example usecase:
```ruby
# /spec/requests/controller_spec.rb
response 400, "bad request" do
  schema { "$ref" => "#/components/responses/400" }

  let(:params) { { key: 'bad-param' } }
  run_test!
end


# /components/responses/400.json
{
  "description": "Response body for a bad request",
  "type": "object",
  "properties": {
    "errors": {
      "type": "array",
      "items": {
        "type": "string",
      }
    }
  }
}
```

### DIFF
--- a/rswag-specs/lib/rswag/specs/response_validator.rb
+++ b/rswag-specs/lib/rswag/specs/response_validator.rb
@@ -62,7 +62,19 @@ module Rswag
             swagger_doc.slice(:definitions)
           else
             components = swagger_doc[:components] || {}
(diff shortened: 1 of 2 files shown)


tyler-wel:
This was a misunderstanding of how response schemas work, closing

PR #369, opened 2020-10-17
CI via Github Actions

jamie:
Travis stopped builds around April, this Github Actions support brings it in-house.

The sole caveat: I've had to disable the one javascript-dependent test verifying rswag-ui is loading the api description properly. Opened an issue to track that.

Also sets up a build matrix to run against Ruby 2.6 and 2.7, for Rails 5.2 and 6.0.


jamie:
Merging this in to get CI support for future PRs, no functional changes.

PR #342, opened 2020-07-02
[Feature] Add requestBody.required flag

bspellacy:
## Context
* Addresses #341 
* Adds the ability for users to set [the requestBody](https://swagger.io/docs/specification/describing-request-body/) as required for endpoints that consumer either a `body` or `formData`

## Usage
```ruby
parameters name: :item, in: :body, required: true, schema: { '$ref' => '#/components/schemas/item' }
```

PR #371, opened 2020-10-17
PR Grooming

### LINKED ISSUE #325: OAS 3 Document has examples in the wrong path

### LINKED ISSUE #341: [Feature Request] Support for requestBody being required
# Problem
Open API 3.x adds support for `requestBody` to be required. By default, it is optional. Currently, rswag doesn't have the option to declare the request body as mandatory.

# Proposed Solution
Add support for a `required` key to the `parameter` method. If the parameter is `in: :body`, the `required` boolean will be respected. E.g.
```ruby
parameter name: :order, in: :body, required: true, schema: { '$ref' => '#/components/schemas/order' }
```
I believe modifying this [method](https://github.com/rswag/rswag/blob/bec4f16676c6bd7b5ef14cf0dc340a6525429567/rswag-specs/lib/rswag/specs/swagger_formatter.rb#L142) should be enough to get it done. I'd be happy to open up a PR once the team okays the approach.

### LINKED ISSUE #346: Describing Empty Response Body

jamie:
This is a bit of a meta-PR, to help me tag pulls I've reviewed and consider an overall benefit. Will list them here so if this merges to master, Github should close them appropriately. Will also keep changelog updated to reflect those merges.

- resolves #330, resolves #325 response examples
- resolves #347, resolves #346 empty content
- resolves #372 spec deprecation warnings
- resolves #373 extra spec coverage
- resolves #342, resolves #341 requestBody flag for required


jamie:
Calling it an afternoon now, build so far is looking stable.

ramytamer:
@jamie Do you have a rough estimate on when are we gonna release version `2.3.2` with the mentioned PRs?

wwahammy:
Hey all, thanks for the amazing work here. 

Was there a sense of when this might be merged and if there's anything we can do to get it across the finish line?

PR #373, opened 2020-10-17
Output specs

jamie:
Starts setting up some structure for targeted assertions on OpenApi output, given arbitrary rswag.


rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -118,7 +118,8 @@ def add_path(request, metadata, swagger_doc, parameters, example)
 
       def build_query_string_part(param, value)
         name = param[:name]
-        return "#{name}=#{value}" unless param[:type].to_sym == :array
+        type = param[:type] || param.dig(:schema, :type)
```
BookOfGreg, line 121: I see this same change in PR #371 
Could you expand on where `param.dig(:schema, :type)` would come from? I'm trying to understand if this is fixing a bug or adding a new behavior.

PR #319, opened 2020-05-19
Allow OpenAPI 3.0 array query parameter schema (#317)

diachini:
In OpenAPI 3.0, a query parameter can be specified like:

```
name: id
in: query
schema:
  type: array
  items:
    type: string
```
see: https://swagger.io/specification/#parameter-object

In order to support generating the correct OpenAPI 3.0 specification and
to still allow rswag specs to pass, this checks to see if `array` is
specified in the `type` of the parameter, or if it's specified in the
`schema: { type:` instead.

re: #317

PR #388, opened 2020-12-21
Allow only schema type specification on parameters

hrakotobe:
Fixes error on swagger generation when param type is only specified through the schema:
```
     Failure/Error: return "#{name}=#{value}" unless param[:type].to_sym == :array
     
     NoMethodError:
       undefined method `to_sym' for nil:NilClass
     # /Users/hrakotobe/.rvm/gems/ruby-2.6.6/gems/rswag-specs-2.3.1/lib/rswag/specs/request_factory.rb:121:in `build_query_string_part'
```

Sample parameter definition:
```
      parameter name: :myparam,
        in: :query,
        schema: { type: :string },
        required: false
```

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -118,7 +118,8 @@ module Rswag
 
       def build_query_string_part(param, value)
         name = param[:name]
-        return "#{name}=#{value}" unless param[:type].to_sym == :array
+        type = param[:type] || param.dig(:schema, :type)
+        return "#{name}=#{value}" unless type.to_sym == :array
 
         case param[:collectionFormat]
         when :ssv


donny741:
I have encountered the error related to this. When swaggerizing schema with `--dry-run` option turned off, `SwaggerFormatter` deletes `type` key from parameter hash and nests it inside `schema/type` key after the first example. Same mutated parameters array is used for every subsequent example.

Parameters in the first example:
```ruby
[{:name=>:id, :in=>:query, :type=>:id, :required=>true}]
```
Parameters in every other example:
```ruby
[{:name=>:id, :in=>:query, :required=>true, :schema=>{:type=>:string}}]
```

This should fix or at least hide this issue

hrakotobe:
Sorry for the late answer.

I checked again on the current master (`83567e0ee24dfd77fff96450e3b3031c7da81ec4`) and I need both mentions to avoid the error:
```
      parameter name: :query,
        in: :query,
        type: :string,
        schema: { type: :string },
        required: false
``` 

Which generates this JSON in the swagger
```
"parameters": [
          {
            "name": "query",
            "in": "query",
            "type": "string",
            "schema": {
              "type": "string"
            },
            "required": false
          }
```

For me both the inline version and the schema version of type specifications fail when specified alone. Looking at the other PRs, it seems other people are running in the same issue (for instance https://github.com/rswag/rswag/pull/319, if you look at the example, `type` param is duplicated)

I run my project with a modified fork, but I think it would nice to have the mainstream project cleaned as well.

oblakeerickson:
Thank you for this. Looks like I just merged in a similar fix in #392.

Issue #320, opened 2020-05-19
Parameters always empty


heaven:
```ruby
require 'swagger_helper'

describe 'Golfer Registration API', swagger_doc: 'v1/golfer/swagger.yaml' do
  path '/registration' do
    post 'Creates a Golfer record' do
      tags 'Golfer Registrations'

      parameter name: :registration, in: :body, schema: {
        '$ref' => '#/components/schemas/registration'
      }

      let(:registration) do
        {
          email:                 Faker::Internet.email,
          first_name:            Faker::Name.first_name,
          last_name:             Faker::Name.last_name,
          phone:                 Faker::PhoneNumber.phone_number,
          password:              "Test1234",
          password_confirmation: "Test1234"
        }
      end

      response '201', 'Created' do
        schema '$ref' => '#/components/schemas/user_access_token'

        run_test! do |response|
          expect(Golfer.count).to eq 1
          expect(Golfer.last.access_tokens.last.key).to eq parsed(response)[:user_access_token][:key]
        end
      end
    end
  end
end
```
Produces:
```yaml
paths:
  "/registration":
    post:
      summary: Creates a Golfer record
      tags:
      - Golfer Registrations
      parameters: []
      responses:
        '201':
          description: Created
          content:
            application/json:
              schema:
                "$ref": "#/components/schemas/user_access_token"
```

Parameters empty even though there's `parameter name: :registration`.

heaven:
It doesn't work with both `yaml` and `json` formats even though specs do pass.

heaven:
Probably a duplicate of #290

heaven:
Here's the bug! https://github.com/rswag/rswag/blob/7ceedab4cb7788ab6b86b96bd2d8f2419e11ebbb/rswag-specs/lib/rswag/specs/swagger_formatter.rb#L67

Commented out this line and parameters now appear in the UI.

This commit https://github.com/rswag/rswag/commit/2af7c13e59a1bacccdd8136db93894acf0e71dac

Should I use anything else instead of `in: :body`?

deXterbed:
@heaven what version you are on? v2.3.1 works fine for me

heaven:
I am on 2.3.1 and it doesn't work. The problem is in the UI, when you click try it out -> execute:
<img width="1155" alt="Screen shot 2021-01-22 at 10 59 56" src="https://user-images.githubusercontent.com/126732/105469487-fb906880-5ca0-11eb-9ba5-e2a8cc10adde.png">

[...]

PR #391, opened 2021-01-22
Inherit consumes from swagger schema. Addresses #320.

heaven:
Currently, every example should specify the `consumes` attribute. If it's missing this leads to a problem described in #320.

To avoid this we can inherit the default `consumes` attribute defined in `swagger_helper` with the ability to overwrite it in every example.

### DIFF
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -56,7 +56,7 @@ module Rswag
                 is_hash = value.is_a?(Hash)
                 if is_hash && value.dig(:parameters)
                   schema_param = value.dig(:parameters)&.find { |p| (p[:in] == :body || p[:in] == :formData) && p[:schema] }
-                  mime_list = value.dig(:consumes)
+                  mime_list = value.dig(:consumes) || doc[:consumes]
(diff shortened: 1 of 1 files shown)


oblakeerickson:
Thank you for this!

I think this is okay, because I think it is good to reduce boilerplate code, but there are a couple of things to be aware of. It appears that the global "consumes" and "produces" blocks are part of [swagger 2](https://swagger.io/docs/specification/2-0/basic-structure/) but not part of [OAS 3](https://spec.openapis.org/oas/v3.0.3#request-body-object). OAS 3 wants the `content` block to be inside of the individual request/response blocks, and doesn't appear to have a global option. 

Your fix works and correctly makes valid OAS3 content blocks for each request, but as a follow up we should remove the global "consumes" and "produces" blocks that are generated since those aren't valid OAS3 (but I think we have a lot of those types of things to clean up in the code base).

Closes #320

heaven:
Thank you. Might be worth adding an option to RSwag configuration then, e.g. when defining separate APIs (swagger.yml docs). In other words any way that’d allow to avoid this duplicate code in every example.

Additionally, if no global setting was set might be worth showing an error or a warning, because currently it’s not obvious that this part is required in every example. I think many users may face this same same problem.

Best,
Alex

PR #392, opened 2021-01-26
don't clobber response content

pfilipow:
Open API 3.0.3 requires example responses to be placed under `response > content > mimetype > example`
However, current rswag deletes everything under the `content` key in order to initialize an empty object
This PR skips content empty object initialization if content is not null, in order to enable the following:

```
after do |example|
    example.metadata[:response][:content] = { 'application/json' => { example: JSON.parse(response.body, symbolize_names: true) } }
end
```

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -118,7 +118,7 @@ module Rswag
 
       def build_query_string_part(param, value)
         name = param[:name]
-        return "#{name}=#{value}" unless param[:type].to_sym == :array
+        return "#{name}=#{value}" unless param[:type]&.to_sym == :array
 
         case param[:collectionFormat]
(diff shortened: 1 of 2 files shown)


rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -118,7 +118,7 @@ def add_path(request, metadata, swagger_doc, parameters, example)
 
       def build_query_string_part(param, value)
         name = param[:name]
-        return "#{name}=#{value}" unless param[:type].to_sym == :array
+        return "#{name}=#{value}" unless param[:type]&.to_sym == :array
```
oblakeerickson, line 121: Ha nice! I just ran into this today.

oblakeerickson (approved):

oblakeerickson:
Thanks!

PR #394, opened 2021-01-29
Fix response examples

donny741:
## Problem
Response examples that are inside `target_node[:content]` attribute are overridden with a blank hash.

## Solution
Seems like `target_node.merge!(content: {})` line was missed during PR merge. Removing it and adding a test case.

### Checklist
- [x] Added tests
- [ ] Changelog updated


oblakeerickson:
Awesome thanks!

PR #396, opened 2021-02-11
drop using #send

kwstannard:
`#send` is dangerous to call on objects passed into your library.

Instead, this will take a page from normal RSpec request specs and allow
developers to specify a "request_params" hash and a "request_headers" hash in let blocks
that this lib can then access.

I had to use request_ because example#headers already exists in request
specs as a shorthand for integration_session.headers.

I think this also cleans up some of the logic nicely.

# RequestFactory initializer

I moved extracting the params & headers into initializer to avoid
duplicating effort.

# Warning

This will obviously break peoples' pre-existing specs, and needs to be
part of a major version change.

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_helpers.rb
@@ -7,7 +7,7 @@ module Rswag
   module Specs
     module ExampleHelpers
       def submit_request(metadata)
-        request = RequestFactory.new.build_request(metadata, self)
+        request = RequestFactory.new(metadata, self).build_request
 
         if RAILS_VERSION < 5
           send(
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -39,7 +39,7 @@ module Rswag
         (operation_params + path_item_params + security_params)
           .map { |p| p['$ref'] ? resolve_parameter(p['$ref'], swagger_doc) : p }
           .uniq { |p| p[:name] }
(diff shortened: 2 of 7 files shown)


jtannas:
I like the idea! Definitely something that's going to take some consideration. I've referenced it in #503 as something that could be part of a v3 release.

romanblanco:
@kwstannard I'd love to get this into v3 release.

I have tried to go ahead and rebase this on top of [v3](https://github.com/rswag/rswag/tree/v3) branch. Mostly the conflits come from https://github.com/rswag/rswag/pull/605 and https://github.com/rswag/rswag/pull/574.

I have the result here: https://github.com/rswag/rswag/compare/v3...romanblanco:rebased-396-2. Would you, please, take a look?

kwstannard:
The rebase looks good! Thank you for handling that.

romanblanco:
@kwstannard Thank you for checking it so quick!

romanblanco:
Reopened the PR, since changes are breaking changes in #605

kwstannard:
I vote reverting 605 in the v3 branch. It is extra code for a problem solved already in this PR.

romanblanco:
@kwstannard, Thanks! I've updated https://github.com/rswag/rswag/compare/v3...romanblanco:rebased-396-2 (and plan to merge it into `v3`), basically reverting the #605 as you suggested in https://github.com/rswag/rswag/commit/ac25323643628eddac1c9d026c5c948c8c747259.

I've kept the specs to show that problem addressed by #605 is still covered. Now, only some documentation covering `request_params` and `request_headers` is missing.

PR #397, opened 2021-02-11
Using Servers block first url during request building for OAS3 compatible documents

developingchris:
This PR adds support for using the servers block during test time to keep from having a v2 test configuration and v3 publish difference.

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -127,7 +127,7 @@ module Rswag
       def base_path_from_servers(swagger_doc)
         return '' unless swagger_doc.has_key?(:servers)
 
-        server = swagger_doc[:servers].firstt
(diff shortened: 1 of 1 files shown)


developingchris:
@oblakeerickson  I'm looking for some initial feedback on this approach? Does this make sense?

kwstannard:
This is an interesting approach. My suggestion would be to set base_path as possibly a Rswag::Specs.config method that defaults to your implementation if not specified.

developingchris:
> This is an interesting approach. My suggestion would be to set base_path as possibly a Rswag::Specs.config method that defaults to your implementation if not specified.

In trying to get our swagger file to validate using several OAS 3 checkers, we received errors when specifying OAS 3 and including the base_path. So while ergonomically that feels good for users of Rswag, it is generating a file which generates errors elsewhere.

kwstannard:
The base_path would be in the config, not in the swagger doc.

PR #405, opened 2021-02-26
Add Truffleruby head to CI

gogainda:


oblakeerickson:
Interesting... I don't know bunch about truffleruby, are there other open source projects you can point to that are also adding this to their CI?

gogainda:
@oblakeerickson  actually bunch of gems already supports truffleruby https://github.com/eregon/truffleruby-gem-tracker/blob/master/gems.yml

PR #406, opened 2021-03-04
Show the response body for comparison when schema checks fail

mynnx:
When schema validation fails, we get information about why the body did not match the schema, but it's not very useful unless we know the _contents_ of the body itself.  This PR adds the response body to schema validation errors, similar to how it's done in the `validate_code!` method in this same file.


oblakeerickson:
Thanks!

PR #410, opened 2021-03-13
Only write YAML for tests that ran

mynnx:
Our team has a separate Swagger doc that we use just for proposals of API changes.  We put those in different spec files with different `swagger_document:` values.  Those tests (for obvious reasons) won't pass yet, but we want to generate API docs for them:

* On the "real" API doc run, we want to run with `--dry-run: false` so we get fully-baked schema examples
* On the "proposal" API doc run, we just want the schema as-written without running tests

So we have to run two mutually-exclusive commands - no big deal, right?  Unfortunately as written, `rswag` writes *all* Swagger documents to disk and not just the ones affected by the tests which were run.  This PR makes it so that the only YAML written is the YAML for the specs which were run.

### DIFF
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -17,7 +17,7 @@ module Rswag
       def initialize(output, config = Rswag::Specs.config)
         @output = output
         @config = config
-        @swagger_docs_affected = Set.new
+        @swagger_docs_affected = []
 
         @output.puts 'Generating Swagger docs ...'
       end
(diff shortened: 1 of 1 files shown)


rswag-specs/lib/rswag/specs/swagger_formatter.rb
```diff
@@ -34,6 +36,7 @@ def example_group_finished(notification)
         return unless metadata.key?(:response)
 
         swagger_doc = @config.get_swagger_doc(metadata[:swagger_doc])
```
mynnx, line 38: I'm not sure how this will behave if you don't define a `swagger_doc` in your test, but we didn't have any in our codebase so I wasn't able to test it out under this circumstance.

Also, if you have suggestions for tests we could write for this change, let me know!

mynnx: Dang, it appears the tests caught something in my code but I'm at a loss as to why it doesn't work.  If a maintainer has a suggestion I'd be happy to investigate.

sergioisidoro:
Indeed this might have some problem with defaults that don't specify the swagger doc.

May I suggest that you use a different API version for these proposals (https://github.com/rswag/rswag#supporting-multiple-versions-of-api) and maybe define in that global metadata attribute if docs should be generated or not for that version?

PR #411, opened 2021-03-16
URL-encode query params before sending them to the controller

mynnx:
Our `rswag` specs used a query param with a `+` symbol as the value in a test.  Once this spec had made its way into the controller, the `+` had been converted into a space character because it already assumed it was URL-encoded.

I'm not sure what the expectations are for query param examples - should we presume the user has already URL-encoded it?

This PR presumes that the user is providing a raw query param value as though someone had typed it into an address bar - this might be a breaking change for people who have come to rely on this behavior.


mynnx:
This appears to no longer be an issue in our code, mysteriously.  It's possible our team updated an upstream dependency, though.

sergioisidoro:
@mynnx is it ok to close this one then?

PR #426, opened 2021-05-11
Update Contributing Doc with Bundle Install Troubleshooting

smridge:
While attempting to make changes and add/update specs, I ran into some issues when running `bundle` using MacOS Big Sur. Afterwards I was able to run `bundle exec rspec spec` successfully. Thought it might be a good one to add to help aid and encourage development.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,24 @@ npm install
 cd -
 ```
 
+### Troubleshooting:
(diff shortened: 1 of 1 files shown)


oblakeerickson (changes requested):
Thanks for working on this. Can you indicate that these are for MacOS Catalina and Big Sur.

smridge:
> Thanks for working on this. Can you indicate that these are for MacOS Catalina and Big Sur.

Good call. Done!

PR #441, opened 2021-08-09
Better error message for missing let

jaydorsey:
I've run into this a number of times, and helped developers triage this
as well. This is meant to improve the developer experience by providing
an improved error message when we forget to configure a let

The original error message is actually a `NoMethodError`/`undefined method`
and it's not always immediately clear what the solution is, particularly if 
you're new to rspec/rswag

Here's an example error (the real one that prompted me to create this PR)

image

And the new error message:

image

I have a number of other small changes like this that I can contribute as well
if these types of pull requests are welcome.

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -194,11 +194,35 @@ module Rswag
 
       def build_json_payload(parameters, example)
         body_param = parameters.select { |p| p[:in] == :body }.first
-        body_param ? example.send(body_param[:name]).to_json : nil
+
+        return nil unless body_param
+
+        raise(MissingParameterError, body_param[:name]) unless example.respond_to?(body_param[:name])
+
+        example.send(body_param[:name]).to_json
       end
 
       def doc_version(doc)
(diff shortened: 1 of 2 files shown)


rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -194,12 +194,36 @@ def build_form_payload(parameters, example)
...
+    class MissingParameterError < StandardError
+      attr_reader :body_param
+
+      def initialize(body_param)
+        @body_param = body_param
+      end
+
+      def message
+        <<~MSG
+          Missing parameter #{body_param}
```
BookOfGreg:
```suggestion
          Missing parameter '#{body_param}'
```

jaydorsey: I updated with this suggestion & rebased off of a more recent master branch to fix the failing specs

jaydorsey:
@BookOfGreg I updated with both fo your suggestions, and fixed the failing specs by doing a rebase/update. Let me know if you want any additional changes

inverse (approved):

PR #442, opened 2021-08-10
Replace therubyracer with mini_racer

jaydorsey:
therubyracer doesn't appear to be under active development

`bundle install` fails because of it; at best it requires some special configuration to run (documented in the contributing document). The contributing guide docs don't appear to work with M1 processor; mini_racer works just fine with both.

Submitting this because I think it can help minimize the barrier to entry to contributing to the repository

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,24 +29,6 @@ npm install
 cd -
 ```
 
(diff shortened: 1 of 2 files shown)


oblakeerickson:
Sweet thanks!

PR #451, opened 2021-09-29
Fix request building with formData parameters

### LINKED ISSUE #348: Upgrade to OpenApi 3.0 gives issues with multipart/form-data
Hi everyone!

While trying to upgrade my swagger definition from Swagger 2 to OpenApi 3.0, I found the following issue with `multipart/form-data` input parameter:

These params in my spec file:
```
consumes 'multipart/form-data'
produces 'application/json'
parameter name: :file, in: :formData, type: :file
parameter name: :kind, in: :formData, type: :string
```

is expected to generate (https://swagger.io/docs/specification/describing-request-body/multipart-requests):
```
"requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "type": "object",
                "properties": {
                  "file": {
                    "type": "string",
                    "format": "binary"
                  },
                  "kind": {
                    "type": "string"
                  }
                }
              }
            }
          }
        }
```
but it generates:
```
        "requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "type": "string"
              }
            }
          }
        }
```

Even the test-app (https://github.com/rswag/rswag/blob/master/test-app/swagger/v1/swagger.json) generated similar `swagger.json` for `/blogs/{id}/upload` which goes against the OpenApi 3 multipart requests specification.

donny741:
## Problem
Swaggerization part and request building part of `formData` parameters behave differently. With [this commit](https://github.com/rswag/rswag/commit/2af7c13e59a1bacccdd8136db93894acf0e71dac) `formData` parameters are being filtered out and only the first one's `:schema` definition is used inside endpoint description (same as with `body` parameters). However, in `request_factory` every single `formData` param is used to build a payload.
So with parameter definition like this:
```ruby
parameter name: :name, in: :formData, type: :string

let(:name) { 'Name' }
```
An empty `requestBody` definition for that API action is generated  (because there is no `:schema` key)

With this variation:
```ruby
parameter name: :form_data, in: :formData, schema: { type: :object, properties: { name: { type: :string } } }

let(:form_data) { { name: :name } }
let(:name) { 'Name'}
```
the correct swagger definition is generated, but the controller receives the `name` parameter that is nested inside the `form_data` object. And this does not match the defined schema.

## Solution
This can be fixed by either changing `request_factory` so it puts first `formData` parameter inside request body (implemented in this PR), or changing `swagger_formatter` so that it builds schema definition from multiple `formData` parameters. Don't know the reason why `swagger_formatter` was initially implemented this way, so I went with the first option.

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -253,10 +253,8 @@ module Rswag
         # Rather that serializing with the appropriate encoding (e.g. multipart/form-data),
         # Rails test infrastructure allows us to send the values directly as a hash
         # PROS: simple to implement, CONS: serialization/deserialization is bypassed in test
-        tuples = parameters
-          .select { |p| p[:in] == :formData }
-          .map { |p| [p[:name], example.send(p[:name])] }
-        Hash[tuples]
+        form_data_param = parameters.select { |p| p[:in] == :formData }.first
+        form_data_param ? example.send(form_data_param[:name]) : {}
       end
 
       def build_json_payload(parameters, example)
--- a/rswag-specs/spec/rswag/specs/request_factory_spec.rb
+++ b/rswag-specs/spec/rswag/specs/request_factory_spec.rb
@@ -344,19 +344,12 @@ module Rswag
           context 'form payload' do
             before do
               metadata[:operation][:consumes] = ['multipart/form-data']
-              metadata[:operation][:parameters] = [
-                { name: 'f1', in: :formData, type: :string },
-                { name: 'f2', in: :formData, type: :string }
-              ]
-              allow(example).to receive(:f1).and_return('foo blah')
-              allow(example).to receive(:f2).and_return('bar blah')
+              metadata[:operation][:parameters] = [{ name: 'comment', in: :formData, schema: { type: 'object' } }]
+              allow(example).to receive(:comment).and_return(text: 'Some comment')
             end
 
-            it 'sets payload to hash of names and example values' do
-              expect(request[:payload]).to eq(
-                'f1' => 'foo blah',
-                'f2' => 'bar blah'
-              )
+            it "sets payload to hash of names from first 'formData' parameter" do
+              expect(request[:payload]).to eq(text: 'Some comment')
             end
           end
         end
--- a/test-app/spec/integration/blogs_spec.rb
+++ b/test-app/spec/integration/blogs_spec.rb
@@ -141,15 +141,19 @@ RSpec.describe 'Blogs API', type: :request, swagger_doc: 'v1/swagger.json' do
       operationId 'uploadThumbnailBlog'
       consumes 'multipart/form-data'
       parameter(
-        name: :file,
+        name: :form_data,
         description: "The content of the blog thumbnail",
         in: :formData,
-        type: :file,
+        type: :object,
+        properties: {
+          file: { type: :file },
+        },
         required: true
       )
 
       response '200', 'blog updated' do
-        let(:file) { Rack::Test::UploadedFile.new(Rails.root.join("spec/fixtures/thumbnail.png")) }
+        let(:form_data) { { file: file } }
+        let(:file) { Rack::Test::UploadedFile.new(Rails.root.join('spec/fixtures/thumbnail.png')) }
         run_test!
       end
     end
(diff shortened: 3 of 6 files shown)


rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -186,10 +186,8 @@ def build_form_payload(parameters, example)
         # Rather that serializing with the appropriate encoding (e.g. multipart/form-data),
         # Rails test infrastructure allows us to send the values directly as a hash
         # PROS: simple to implement, CONS: serialization/deserialization is bypassed in test
-        tuples = parameters
-          .select { |p| p[:in] == :formData }
-          .map { |p| [p[:name], example.send(p[:name])] }
-        Hash[tuples]
+        form_data_param = parameters.select { |p| p[:in] == :formData }.first
```
sergioisidoro, line 256: That `.first` worries me. Doesn't this just work with the 1st form parameter? 
```
{ name: 'f1', in: :formData, type: :string },
{ name: 'f2', in: :formData, type: :string }
```

Wouldn't `f2` be excluded?

The test seemed to test for 2 parameters in FormData, but now it only tests one. Could you return the test to have 2 form parameters to make sure it works with more that 1 form parameter?

sergioisidoro (changes requested):

rswag-specs/lib/rswag/specs/request_factory.rb
donny741: @sergioisidoro Thanks for your feedback! Yes, that's exactly how it would work. I have explained the reasoning in the "solution" paragraph. `body` parameters [work the same way](https://github.com/rswag/rswag/blob/master/rswag-specs/lib/rswag/specs/request_factory.rb#L235).
Another solution would be to make [swagger_formatter](https://github.com/rswag/rswag/blob/master/rswag-specs/lib/rswag/specs/swagger_formatter.rb#L58) accept multiple parameters because now it accepts only the first one and logic in these places diverge.

I agree this solution isn't great and is more of a quick fix. Currently, to work around this issue we define the first formData param just for `SwaggerFormatter` and then redefine the same schema in separate params to make `RequestFactory` work.

sergioisidoro: Ok, let me take a bit longer to look into this, and see if I can figure out the historical background behind the body param. My guess is that it is also not supposed to be like that 🤔

mlh758: I don't think the current implementation is actually working for people. #348 is a long thread of people resigned to either having the integration test work, or their generated documentation being correct. I was about to submit pretty much this same PR but happy to see it's already been done.

I also think that conceptually it makes sense to only have a single body parameter supported since you can only send one body.

kennon:
+1 :-)

abstracts33d:
+1 :-)

rswag-specs/lib/rswag/specs/request_factory.rb
braindeaf: I have come up against this issue recently. And I am very keen to solve it, happy to work on it if we get more ideas on the best solution.

nparker-tc:
I have run into this issue recently and the workaround here: https://github.com/rswag/rswag/issues/348#issuecomment-1386130773 worked for me. 

```ruby
 parameter name: :'', in: :formData, schema: {
  type: :object,
  properties: {
    client_id: { type: :string },
    client_secret: { type: :string },
    grant_type: { example: 'client_credentials', type: :string },
  },
  required: [ 'client_id', 'client_secret', 'grant_type' ]
}

response '200', 'Token created' do
  example 'application/json', :succesful_token_example, {
    'access_token': 'FeXebuMLItU3UPpcXlUgGkxeXoWLgX9CULUQ6FBAFjs',
    'token_type': 'Bearer',
    'expires_in': 7200,
    'scope': 'public',
    'created_at': 1677685057
  }
  let(:'') { { client_id: 'client_id', client_secret: 'client_secret', grant_type: 'client_credentials' } }
  run_test!
end
```
image

mlh758:
I never thought I'd say this... but I think I prefer the monkey patch 😞

braindeaf:
Which solution was the monkey patch?

Drowze:
Just faced this issue today - I don't think there's currently a good way to document `formData` parameters using OAS 3.0 🤔 

Looking forward to this change :)

MichaelAlexanderBanuelos:
any update on this?

Sean0628:
Hi, just following up to see if there’s been any update on this issue. Let me know if there’s anything I can help with! Confirmed, [this workaround](https://github.com/rswag/rswag/issues/348#issuecomment-1386130773) works for now.

Flixt:
This is still a problem

pgiblock:
Just hit this 4 year old bug. Can the PR please be merged in?

PR #453, opened 2021-10-15
Update swagger-ui to 3.52.5

jswk:


inverse (approved):

inverse:
@oblakeerickson any chance we could get this merged and a new build shipped? 

Was getting the following JS errors previously

image


When it was trying to tender a response like this. The problem was the `oneOf` with `array` and `string` types.

```yml
    error_response:
      type: object
      properties:
        status:
          type: integer
          example: 422
        title:
          type: string
          example: InvalidParamsExeption
        details:
          oneOf:
          - type: array
            items:
              type: object
          - type: string
 ```

PR #455, opened 2021-10-27
add custom error in case user forgets to define name with let

SteveRedka:
If you declare a name parameter, but forget to define it within `it` or `let` group, RSpec will raise undescriptive error. Two things can happen:
1. Usually it will raise `NoMethodError`. Which is mostly true, but it's very generic and doesn't say anything about the place where this error happened and how to fix it.
```ruby
undefined method `blog_id' for #<RSpec::ExampleGroups::ApiTrackers::ApiTrackersIdd::Get::Successful:0x000055e89bf1dab8>
```
2. Some variables like `id` can raise `WrongScopeError`
```ruby
`id` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
```
The text of this error is just wrong and sends user on wild goose chase. 

 This pull request adds custom error to remind user to define the missing variable.

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -106,6 +106,10 @@ module Rswag
 
         request[:path] = template.tap do |path_template|
           parameters.select { |p| p[:in] == :path }.each do |p|
+            unless example.respond_to?(p[:name])
+              raise ArgumentError.new("`#{p[:name].to_s}` parameter key present, but not defined within example group"\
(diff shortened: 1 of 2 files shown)


jtannas:
I like it! I've definiely chased that wild goose a couple times and this would've saved a ton of time :swan: 
This is also applicable to query parameters with `required: true` if you feel like extending the PR a bit. If not we can just merge as-is.

BookOfGreg (approved):

PR #470, opened 2022-01-10
Fix github actions to use a matching Ruby version 

ziyunli:
Currently, the github actions always use Ruby 2.6 even though there are a couple of other versions specified. For example https://github.com/rswag/rswag/runs/4413833399?check_suite_focus=true.

image

This fix here will use a matching Ruby version based on the matrix, Following the setup-ruby action doc https://github.com/ruby/setup-ruby#matrix-of-ruby-versions

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -19,7 +19,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: ruby/setup-ruby@v1
-      with: { ruby-version: 2.6 }
+      with:
+        ruby-version: ${{ matrix.ruby }}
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Verified, thank you @ziyunli !
This could be one of the things that was messing with adding Rails 7 to the test suite since ruby 2.6 isn't supported on that version of Rails.
https://github.com/rswag/rswag/runs/4912391570?check_suite_focus=true#step:3:3

BookOfGreg:
Had to cancel the workflow run as truffleruby-head took over 10 minutes to build and looked to have hung.
Will merge this and clean up whatever is happening to Truffleruby. Thank you again.

jtannas:
@BookOfGreg still want truffleruby added to the test matrix?

PR #471, opened 2022-01-13
Use File.exist? instead of File.exists?

benjaminwols:
It seems Ruby 3.1 will remove File.exists? which has been obsolete for quite some time.
This fixes the deprecation warning: `rswag-ui-2.4.0/lib/rswag/ui/middleware.rb:40: warning: File.exists? is deprecated; use File.exist? instead`
File.exist? already exists since ruby 1.8.7: https://ruby-doc.org/core-1.8.7/File.html#method-c-exist-3F

### DIFF
--- a/rswag-ui/lib/rswag/ui/middleware.rb
+++ b/rswag-ui/lib/rswag/ui/middleware.rb
@@ -37,7 +37,7 @@ module Rswag
       end
 
       def template_filename
-        @config.template_locations.find { |filename| File.exists?(filename) }
+        @config.template_locations.find { |filename| File.exist?(filename) }
       end
     end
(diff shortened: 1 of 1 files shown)


sergioisidoro (approved):

sergioisidoro:
Looks good. The fact that the tests are not failing in master, now that we added Ruby 3.1 in CI, indicates our coverage is not ideal. If anyone wants to contribute to a test case on this it would be appreciated :)

BookOfGreg:
Triggered new run here:
https://github.com/rswag/rswag/actions/runs/1815385532

Edit: All passed. 🟢  Thank you!

PR #472, opened 2022-01-18
keep existing metadata in requestBody.content.mime

fnordfish:
This allows for adding multiple `requestBody` `examples` via metadata.

```ruby
after do |example|
  media_type = request.media_type
  example.metadata[:operation].deep_merge!(requestBody: { content: { media_type => { examples: {} } } })
  example.metadata[:operation][:requestBody][:content][media_type][:examples].tap { |e|
    e[e.size] = {
      summary: example.example_group.description,
      description: example.description,
      value: request.body.string
    }
  }
end
```

I tries to add specs, but didn't quite understand the existing specs. Hope the sample above helps to understand my intent

### DIFF
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -61,7 +61,7 @@ module Rswag
                     value[:requestBody] = { content: {} } unless value.dig(:requestBody, :content)
                     value[:requestBody][:required] = true if schema_param[:required]
                     mime_list.each do |mime|
-                      value[:requestBody][:content][mime] = { schema: schema_param[:schema] }
+                      (value[:requestBody][:content][mime] ||= {})[:schema] = schema_param[:schema]
                     end
                   end
 


vokshirg:
it would be better to add this possibility to core [here](https://github.com/rswag/rswag/blob/master/rswag-specs/lib/rswag/specs/swagger_formatter.rb#L64 )

sergioisidoro:
Hi. If I'm not mistaken, this is already being addressed by various PRs, and we're trying to get those forward. #302 and #305 
Do you think this can be closed?

fnordfish:
If you're asking me, I've no idea. I was looking at open PRs, but couldn't make out if they would do what I wanted. Also they seemed quite old, so I was hesitant trying them out.
This PR was to demonstrate and find a minimal solution to this specific issue.

sergioisidoro:
Ok, then we will keep this open in case the other ones don't move forward.
As you said this seems like a quick solution to the problem that we can later iterate.

sergioisidoro:
@fnordfish If this is about examples, it should be resolved here: https://github.com/rswag/rswag/pull/501

PR #473, opened 2022-01-20
Remove requirement to include invalid 'consumes' attribute

micapam:
I have made this PR in an attempt to fix a problem with rswag's compatibility with OpenAPI v3.

Currently, the only way I can get rswag to produce `requestBody` on a request is to include the attribute 'consumes' on the root of the document. 

This attribute, however, is invalid in OpenAPI v3. I suggest we remove the requirement to include the invalid attribute 'consumes'.


sergioisidoro:
@micapam It seems that according to OAS3 you always need a mime type for the request body:
https://spec.openapis.org/oas/v3.1.0.html#request-body-examples

Can you use `"*/*"` like the example in the OAS3 specs as a 'consumes'?

PR #477, opened 2022-01-22
Add latest ruby and rails

BookOfGreg:
Adds latest Rails patches
Adds Rails 7
Adds Ruby 3.0 and 3.1
Removes Ruby 2.6 and TruffleRuby


sergioisidoro:
Ah, I missed this. I was also working on it:
https://github.com/rswag/rswag/pull/481

We need to add some exclusions to the test matrix due to the Ruby/Rails requirements

PR #479, opened 2022-01-23
Fix truffleruby workflow

BookOfGreg:
using truffleruby-head is supported
https://github.com/ruby/setup-ruby#matrix-of-ruby-versions
However for us it times out. Unclear why.
https://github.com/rswag/rswag/runs/4912391605?check_suite_focus=true
This continues #470

PR #481, opened 2022-01-24
Testing env improvements - add test environments and bump versions

sergioisidoro:
Some housekeeping:

* Bumps patch and minor versions on test matrix
* Adds Rails 7 to test matrix
* Adds Ruby 3.0 and 3.1 ( Excluding older Rails versions from tests in Ruby 3.x )
  * Docs [for excluding envs here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix)
* Adds some annotations about EOLs of testing environments
* Adds a Docker and Docker compose to easily debug in any environment (ARM64 makes it sometimes tricky to use older Ruby versions)
* Removes `truffleruby-head` from the test matrix as it was a bit discussed on the PRs Greg authored. -- eg https://github.com/rswag/rswag/pull/479#issuecomment-1019474574

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,11 @@ when  '6', '7'
   gem 'sqlite3', '~> 1.4.1'
 end
 
+case RUBY_VERSION.split('.').first
+when '3'
+  gem 'net-smtp', require: false
+end
+
 gem 'rswag-api', path: './rswag-api'
 gem 'rswag-ui', path: './rswag-ui'
 
--- a/Gemfile
+++ b/Gemfile
@@ -4,21 +4,16 @@ source 'https://rubygems.org'
 
 # Allow the rails version to come from an ENV setting so Travis can test multiple versions.
 # See http://www.schneems.com/post/50991826838/testing-against-multiple-rails-versions/
-rails_version = ENV['RAILS_VERSION'] || '5.2.4.2'
+rails_version = ENV['RAILS_VERSION'] || '5.2.6'
 
 gem 'rails', rails_version.to_s
 
-case rails_version.split('.').first
-when '3'
-  gem 'strong_parameters'
-when '4', '5', '6'
-  gem 'responders'
-end
+gem 'responders'
 
 case rails_version.split('.').first
-when '3', '4', '5'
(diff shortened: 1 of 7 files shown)


sergioisidoro:
```
            LoadError:
              cannot load such file -- net/smtp
```

Potentially an upstream problem - https://stackoverflow.com/a/70500221/2943164

sergioisidoro:
```
NameError:
  uninitialized constant ActiveSupport::XmlMini::IsolatedExecutionState
```
Some more upstream issues https://github.com/rails/rails/issues/43851

sergioisidoro:
@BookOfGreg At least test suite is now green with Rails 7 🎉
Some significant issues upstream were fount, but have patches, but I think they are not relevant to the gem itself.

Let me know if this is enough to get this forward, and sorry for the duplicate work that you started.

Gemfile
```diff
@@ -4,24 +4,24 @@ source 'https://rubygems.org'
...
-when '3'
-  gem 'strong_parameters'
-when '4', '5', '6'
-  gem 'responders'
-end
+gem 'responders'
 
 case rails_version.split('.').first
-when '3', '4', '5'
+when '5'
```
sergioisidoro, line 14: I kind of deprecated Rails 3 and 4 here.
But as Rails 5 is reaching its end of life I think it's quite reasonable (besides it's not tested)

BookOfGreg: Yes they're out of support by Rails so they're good to drop.

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    image: rswag
+    ports:
+      - 3000:3000
+    build:
+      context: .
+    volumes:
+       - ./test-app:/rswag/test-app
+  test:
+    <<: *base-test
+    command: sh -c 'bundle exec rake db:setup && bundle exec rspec'
```
BookOfGreg, line 14: Good move for local convenience :+1:

BookOfGreg (approved):
Thanks for repairing and adding to the test suite here.

Issue #490, opened 2022-02-09
Maintainer Tooling Discussion


jtannas:
Making a thread here for discussing different things we can do to make maintaining the project less time consuming.

Ideas:
- Add a `SECURITY.md` file for reporting vulnerabilities to head off tickets like #135
- Add a `rubocop.yml` to keep everyone on the same page for project styling; lint & autocorrect PRs
- Add in spellchecker ([cspell](https://cspell.org/)?) for PRs
- Add https://github.com/marketplace/stale to help keep stale issues under control
- Are the issue templates working? I don't see anything come up when creating a new template. Assuming they're broken:
  - Add in a template chooser https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
  - Create issue forms (or wait until beta is complete?) https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms
- Set up a project wiki with an FAQ page; Split the README into wiki pages
- Set up GitHub Discussions so that we can focus the "Issues" section on concrete issues
- ???

BookOfGreg:
- [x]  #491  - We need to discuss the content of this file, generally it's expected that there's an email to handle this privately, however given the length of time it took me to get back in the game, what kind of guaranteed can we make?

BookOfGreg:
- [x] #492 - This one should be fairly straight forward. Stale is pretty sensible to me, shame to loose people's good content but there is a limit of what's useful and what will reasonably be gotten to. Discuss in the PR thresholds and message.

BookOfGreg:
- [x] I've made a discussion section. I've never used one before but happy for others to do so. Hopefully this will help us make and keep this a vibrant place.

You can find it here: https://github.com/rswag/rswag/discussions

jtannas:
Thoughts on opening up a wiki?
I'm seeing some common threads in the tickets that would fit well into an FAQ section.
- How can split my `swagger_helper` definitions across multiple files?
- Why is this test passing when there are additional properties?
- Why is this test passing when there are missing properties?
- https://github.com/rswag/rswag/issues/256

They could go into the README but it's getting pretty unwieldy with how long it is

jtannas:
Related: can I please get a `documentation` label for keeping track of things to add to an FAQ?
[...]

PR #491, opened 2022-02-09
Add Security tooling

BookOfGreg:
Addressing #490

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,23 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,23 @@
...
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 2.5.x   | :white_check_mark: |
+| < 2.5.0 | :x:                |
+| < 1.x.x | :x:                |
```
jtannas: With a small team of maintainers on a niche gem, I think it's reasonable to only support security updates on the latest version. Could save us the hassle of updating the table with every release if we just put that in writing.

Issue #467, opened 2021-12-23
Installing from github gives 404 assets


hss-mateus:
As pointed in #465, the latest release of rswag on RubyGems doesn't work with Rails 7, so i pointed this github repo at my gemfile.

The problem is that rswag-ui assets are not installed when using this workaround. This requires me to manually install them at the gem folder, like the example:

```sh
ASSETS_PATH=$(bin/rails r "print Rswag::Ui.config.assets_root")

yarn --cwd $ASSETS_PATH install
```

A possible solution for this is to release a new version to RubyGems, including the installed assets.

agildav:
👀

matiassalles99:
Yes a new release would be great, I'm trying to upgrade to Rails 7 also

anmarchenko:
Same here, would really appreciate new release. Thanks for your work!

strifek:
The workaround could be to download `swagger-ui-bundle.js`, `swagger-ui-standalone-preset.js` and `swagger-ui.css` from `https://registry.npmjs.org/swagger-ui-dist/-/swagger-ui-dist-3.52.5.tgz` and manually place it in your `public` folder (must match your swagger path, eg. `public/api-docs`) to serve them as a static files.

coins-black:
Has the same bug

mxdavis:
👀

joelbiffin:
👀

jlestavel:
For those who would be looking for a temporary solution:

The Github version of the `rswag-ui` gem does not have the `node_modules` folder, so indeed the `Rswag::Ui.config.assets_root` folder does not exist...

https://github.com/rswag/rswag/blob/a813b63bcc64e9ea49f6416ab4fe023b7f090ad8/rswag-ui/lib/rswag/ui/configuration.rb#L22

The reason of that is that the `node_modules` folder is generated on build (with `npm install`) :

https://github.com/rswag/rswag/blob/a813b63bcc64e9ea49f6416ab4fe023b7f090ad8/ci/build.sh#L20-L21

and is kept with the gem, as you can see in commit 334671cb95da8b665e5b3397f995c1326bc607d1. So we have to wait for a new gem release to have the `node_modules` present in the gem.

In the meantime, what you can do is use a local version of those assets. For our app, as we use webpacker, we added `swagger-ui-dist` to our `package.json`. Then, we ask the gem to use our local `node_modules/swagger-ui-dist` instead of the one that is supposed to be embedded in the gem. The config to do that is:

[...]

PR #493, opened 2022-02-10
add npm install to rswag-ui deploy process

BookOfGreg:


BookOfGreg:
Unfortunately to test/run the deploy process it needs to be on the main branch so this is being merged pre-checks.

BookOfGreg:
This fixes the problem in #467 
image

PR #496, opened 2022-02-13
Issue and PR template updates

sergioisidoro:
- While triaging the issues, documentation is quite often missing in Pull requests, so added a bullet a checkbox for updating docs 
- Also it's unclear what version people are targeting with the older versions. This might not be a problem in the future, but it's nice to ask for the contributors to have backwards compatibility in mind
- Added headers and metadata in the templates
- Added a config so Github will display a issue template chooser
- Moved pull request template out of the folder, because right now we only have on pull request template


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
BookOfGreg, line 1: :+1: I assume this is what we were missing for the issue picker.

BookOfGreg (approved):

sergioisidoro:
Wohoo! :)

Screenshot 2022-02-13 at 15 09 53

.github/ISSUE_TEMPLATE/config.yml
sergioisidoro: Yes. I think just the existence of the file will trigger the picker, but since we're here, let's also not allow blank issues :)

PR #499, opened 2022-02-21
Support all possible query string style and format combinations

Flixt:
rswag does currently not support all possible `style`, `format` options in query strings for OAS 3.0.3 (and so not 3.1). And even allows mixing OAS v2 `collectionFormat` with newer `schema` option.

This PR changes this by implementing all combinations for the `style` and `format` options that the OpenAPI specification lists for 3.0.3 and 3.1.

This change should be backwards compatible with OAS 2 but is probably breaking for people using OAS 3.0.3 together with the `collectionFormat` option, which really should only be supported in OAS 2. I'm not sure about this detail in particular, but  from my point of view it is simply not correct to allow people using `collectionFormat` in a OAS 3.x world.

If this change would be welcome, I'd update the changelog etc as needed.

* [OAS 3.0.3 Style values](https://spec.openapis.org/oas/v3.0.3#style-values)
* [OAS 3.1 Style values](https://spec.openapis.org/oas/v3.1.0#style-values)

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Checklist
- [x] Added tests
- [ ] Changelog updated

### DIFF
--- a/rswag-specs/spec/rswag/specs/request_factory_spec.rb
+++ b/rswag-specs/spec/rswag/specs/request_factory_spec.rb
@@ -77,12 +77,14 @@ module Rswag
           end
         end
 
-        context "'query' parameters of type 'array'" do
+        context "swagger 2.0 'query' parameters of type 'array'" do
+          let(:parameter) do
+            { name: 'things', in: :query, type: :array, collectionFormat: collection_format }
+          end
+
           before do
-            metadata[:operation][:parameters] = [
-              { name: 'things', in: :query, type: :array, collectionFormat: collection_format }
-            ]
-            allow(example).to receive(:things).and_return(['foo', 'bar'])
+            metadata[:operation][:parameters] = [parameter]
+            allow(example).to receive(parameter[:name]).and_return(['foo', 'bar'])
           end
 
           context 'collectionFormat = csv' do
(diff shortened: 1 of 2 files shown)


BookOfGreg:
> probably breaking for people using OAS 3.0.3 together with the collectionFormat option, which really should only be supported in OAS 2.

The good thing about building a gem to a spec, if anyone was using collectionFormat in 3.x then that is out-of-spec, and we don't need to consider that backward compatible. 🎉 

Edit: This looks good so far, need more time to review properly however.

Flixt:
This should be cross checked with https://github.com/rswag/rswag/pull/507

akostadinov:
Who should review this?

sergioisidoro:
~~If I'm reading this correctly, this should fix the issue of #507 for OAS3, but not for OAS2. The tests in this PR cover pretty much the same cases as the ones added in #507, but only for OAS3.~~

~~We've been discussing in this thread - https://github.com/rswag/rswag/discussions/503 - that it's increasingly cumbersome to keep OAS2, and so far no one objected to its deprecation~~

~~I personally think that we should issue a deprecation of OAS2, and start working to cleanup the code, and move this PR forward instead of 507 (and accepting that the bug in array param serialisation in OAS2 wont be fixed).~~

@BookOfGreg what do you think?

EDIT: Ok, I completely misread 503, and it's also a 3.x problem.

sergioisidoro (approved):

rswag-specs/spec/rswag/specs/request_factory_spec.rb
```diff
@@ -119,6 +121,101 @@ module Specs
...
+            it 'formats as exploded pipeDelimited' do
+              expect(request[:path]).to eq('/blogs?things=foo|bar')
+            end
+          end
+
+          context 'pipeDelimited explode=true' do
+            let(:style) { :pipeDelimited }
+            let(:explode) { true }
+            it 'formats as exploded pipeDelimited' do
+              expect(request[:path]).to eq('/blogs?things=foo|bar')
```
sergioisidoro, line 197: Should explode `Should explode /blogs?things=foo&things=bar`
See reference

rswag-specs/spec/rswag/specs/request_factory_spec.rb
```diff
@@ -119,6 +121,101 @@ module Specs
...
+            it 'formats as exploded spaceDelimited' do
+              expect(request[:path]).to eq('/blogs?things=foo bar')
+            end
+          end
+
+          context 'spaceDelimited explode=true' do
+            let(:style) { :spaceDelimited }
+            let(:explode) { true }
+            it 'formats as exploded spaceDelimited' do
+              expect(request[:path]).to eq('/blogs?things=foo bar')
```
sergioisidoro, line 181: Should explode `/blogs?things=foo&things=bar`
See reference

rswag-specs/spec/rswag/specs/request_factory_spec.rb
```diff
@@ -119,6 +121,101 @@ module Specs
...
+            it 'formats as exploded form' do
+              expect(request[:path]).to eq('/blogs?things=foo&things=bar')
+            end
+          end
+
+          context 'spaceDelimited explode=false' do
+            let(:style) { :spaceDelimited }
+            let(:explode) { false }
+            it 'formats as exploded spaceDelimited' do
+              expect(request[:path]).to eq('/blogs?things=foo bar')
```
sergioisidoro, line 173: Should probably encode space with `%204%`

sergioisidoro (changes requested):
I hade approved this but after more carful review, there're are some issues here:

Screenshot 2022-04-28 at 15 01 25

Combination of `spaceDelimited` or `pipeDelimited` with `explode` seems to favour the explosion according to the open API spec, which seems to be contrary to this implementation.

sergioisidoro:
I went through https://github.com/rswag/rswag/pull/507 again, and it implements the same functionality as this pr, with the same test coverage, and according to the OAS specs for the explosion and style. I approved and merged the other PR instead, so this issue should be resolved.

Thank you for taking a shot at this one :) Can we close this PR?

Flixt:
Hey,

fair. I wrongly interpreted the examples in https://spec.openapis.org/oas/v3.0.3#style-values as `explode` has no effect on when using `spaceDelimited`, `pipeDelimited` etc.
On the other hand side I think the advantage of this PR would have been, that it raises errors for undefined combinations of `style` and `type` e.g. `spaceDelimited` style for objects (supported in OAS > 3.1). and it splits up the parameter generation into separate methods.

Anyway, I'm happy the functionality is there now, so I'm closing this PR as being obsolete 👯 .
I'd happy to contribute a new PR to refactor the existing code to raise exceptions und split it into methods if you'd consider that valuable, just let me know.

PR #513, opened 2022-04-28
Fix dynamic example generation docs

sergioisidoro:
`example.metadata[:response][:content]` is not always initialised (if there's no other examples). There is a #TODO in the code to move the initialisation of `content` somewhere else, but this is just a quick fix in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,16 +615,19 @@ end
 To enable examples generation from responses add callback above run_test! like:
 
 ```
-after do |example|
-  example.metadata[:response][:content]['application/json'][:examples] = {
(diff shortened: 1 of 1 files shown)


BookOfGreg (approved):
:+1:

romanblanco:
> There is a #TODO in the code

https://github.com/rswag/rswag/commit/7e0289005218002c39e66a464a9ce82525877226#diff-7c43e7d1fb91b5c86d12f8c431b704ea535c00e16694677fda4ecb67b9277fc7R81

Issue #464, opened 2021-12-07
Plans for OA 3.1


CR4567:
I know that this repo is not fully managed anymore. But does anyone has plans to work on support for OA 3.1?
If there will be others, I would also try to do as much as possible.

oblakeerickson:
Yes. I'm on board for getting rswag to support OA 3.1. Any help would be greatly appreciated :)

BookOfGreg:
:+1: Anyone looking to help will also get my attention.

joelbiffin:
I'd be happy to contribute to this

tom-quiltt:
I will be happy to contribute to this. Especially looking to support `New top-level element for describing Webhooks that are registered and managed out of band`.

@BookOfGreg or anyone, is there a list of features that potential contributors can look into?

jtannas:
Heyo - It's mostly @BookOfGreg, @sergioisidoro, and myself trying to revive this repo.
Right now my focus is on getting through the PR backlog before starting on anything major so that we're not losing existing contributions to merge conflicts. I'm moving Monday though so probably not much is going to happen for a week. Help with PR reviews would be really appreciated.

---

My rough plan is:
1. get through the existing PR backlog so that we're not losing existing contributions to merge conflicts
2. Make the repo easier to maintain
    - rubocop autocorrect & enforce via CI to cut down on style PRs & PR comments
    - put in a Cypress E2E test suite for more confidence when making changes
    - drop support for Ruby 2.6 (EOL) and RSpec 2 (v3 was released in 2014)
    - enforce code & branch test coverage in CI? :thinking: 
3. Bugfixes & Features based on community feedback
    - Full OAS 3.0 / 3.1 support seem high on that list
    - Response examples are something that show up frequently in issues
    - ???

romanblanco:
Will leave this here: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

pboling:
> drop support for Ruby 2.6 (EOL) and RSpec 2 (v3 was released in 2014)

Just wanted to clarify - the internal specs of the project are already expecting RSpec 3.  It is support for end-users with RSpec 2 suites that is to be dropped, correct?

pboling:
> rubocop autocorrect & enforce via CI to cut down on style PRs & PR comments

I have a gem for that, which is easy to integrate - https://rubocop-lts.gitlab.io/

> enforce code & branch test coverage in CI? 🤔

[...]

PR #515, opened 2022-05-01
Fix for nullable refs

codebender:
Added a test to https://github.com/rswag/rswag/pull/332 

Not sure if you will agree with where I put the tests

> Previously the nullable check was only being applied to the type
> attribute validation, but schemas using anyOf, allOf, oneOf etc
> were unable to use nullable because they would never be passed
> directly to the field validator. This is crucial when using $ref
> combined with nullable, because these cases require the use of
> allOf.
> 
> The solution here is to simplify the validation extension at the root
> and not to attach the nullable handling to any particular attribute.
> 
> Before this change, the following would fail validation if foo was
> null:
> 
> ```json
> {
>   "type": "object",
>   "properties": {
>     "foo": {
>       "nullable": true,
>       "allOf": [
>         {
>           "$ref": "#/components/schemas/Foo"
>         }
>       ]
>     }
>   }
> }
> ```
> 
> With this change the nullable check is handled at the right spot
> before the allOf is evaluated.

### DIFF
--- a/rswag-specs/spec/rswag/specs/response_validator_spec.rb
+++ b/rswag-specs/spec/rswag/specs/response_validator_spec.rb
@@ -95,6 +95,84 @@ module Rswag
               it 'uses the referenced schema to validate the response body' do
                 expect { call }.to raise_error(/Expected response body/)
               end
+
+              context 'nullable referenced schema' do
+                let(:response) do
+                  OpenStruct.new(
+                    code: '200',
+                    headers: { 'X-Rate-Limit-Limit' => '10' },
+                    body: '{ "blog": null }'
+                  )
+                end
+
+                before do
+                  metadata[:response][:schema] = {
+                    properties: { blog: { '$ref' => '#/components/schema/blog' } },
+                    required: ['blog']
+                  }
+                end
+
+                context 'using x-nullable attribute' do
+                  before do
+                    metadata[:response][:schema][:properties][:blog]['x-nullable'] = true
+                  end
+
+                  context 'response matches metadata' do
+                    it { expect { call }.to_not raise_error }
+                  end
+                end
+
+                context 'using nullable attrtibute' do
+                  before do
+                    metadata[:response][:schema][:properties][:blog]['nullable'] = true
+                  end
+
+                  context 'response matches metadata' do
+                    it { expect { call }.to_not raise_error }
+                  end
+                end
+              end
+
+              context 'nullable oneOf wtih referenced schema' do
+                let(:response) do
+                  OpenStruct.new(
+                    code: '200',
+                    headers: { 'X-Rate-Limit-Limit' => '10' },
+                    body: '{ "blog": null }'
+                  )
+                end
+
+                before do
+                  metadata[:response][:schema] = {
+                    properties: {
+                      blog: {
+                        oneOf: [{ '$ref' => '#/components/schema/blog' }]
+                      }
+                    },
+                    required: ['blog']
+                  }
+                end
+
+                context 'using x-nullable attribute' do
+                  before do
+                    metadata[:response][:schema][:properties][:blog]['x-nullable'] = true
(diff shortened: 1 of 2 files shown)


sergioisidoro:
Thank you! The tests are good for now :)
Upon coser inspection, I'm more afraid that we're splashing the `x-nullable` everywhere where it is not up to spec.

Could something like this work?
```
require 'json-schema'

module Rswag
  module Specs
    class ExtendedSchema < JSON::Schema::Draft4
      def initialize
        super
        @attributes['type'] = ExtendedTypeV4Attribute
        @attributes['allOf'] = ExtendedAllOfAttribute
        @attributes['anyOf'] = ExtendedAnyOfAttribute
        @attributes['oneOf'] = ExtendeOneOfAttribute

        @uri = URI.parse('http://tempuri.org/rswag/specs/extended_schema')
        @names = ['http://tempuri.org/rswag/specs/extended_schema']
      end
    end

    module NullableValidatorMixin
        def self.validate(current_schema, data, fragments, processor, validator, options = {})
            return if data.nil? && (current_schema.schema['nullable'] == true || current_schema.schema['x-nullable'] == true)
            super
        end
    end

    class ExtendedTypeV4Attribute < JSON::Schema::TypeV4Attribute
        prepend NullableValidatorMixin
    end
    
    class ExtendedAllOfAttribute < JSON::Schema::AllOfAttribute
        prepend NullableValidatorMixin
    end
    
    class ExtendedAnyOfAttribute < JSON::Schema::AnyOfAttribute
        prepend NullableValidatorMixin
    end
    
    class ExtendeOneOfAttribute < JSON::Schema::OneOfAttribute
        prepend NullableValidatorMixin
    end

    JSON::Validator.register_validator(ExtendedSchema.new)
  end
end
```

sergioisidoro:
Ok, here are some docs and links to sources that I've seen when validating this. What I mentioned above doesn't work because the `nullable` needs to be on the parent node, not on the ` anyOf, allOf, oneOf`.

This is actually mentioned in the issue description but I didn't understand what it meant. So it is really only the way forward to accept the nullable on every node.

* https://stackoverflow.com/a/48114924/2943164

On upcoming OAS 3.1 this is not a problem because it's fully compatible of JSON schema.
On OAS3.0 this is a know problem. https://github.com/OAI/OpenAPI-Specification/issues/1368

I'm trying to wrap my head around any possible consequences of allowing null values on every type of node, but until we get 3.1 full running, I think we have much choice, since this is something that should be supported by the spec. 

Btw, help wanted to get this to OAS 3.1 https://github.com/rswag/rswag/issues/464

rswag-specs/spec/rswag/specs/response_validator_spec.rb
```diff
@@ -73,6 +73,84 @@ module Specs
             it 'uses the referenced schema to validate the response body' do
               expect { call }.to raise_error(/Expected response body/)
             end
+
+            context 'nullable referenced schema' do
```
sergioisidoro: We could move this to the context bellow `context 'openapi 3.0.1'` because this is actually not supported by OAS2, and we're moving out of OAS2 anyway

sergioisidoro (approved):

rswag-specs/spec/rswag/specs/response_validator_spec.rb
codebender: sure I can move that down in the spec

codebender: @sergioisidoro Ok I moved the new specs down to `context 'openapi 3.0.1' do` and updated the tests to use the `#/components/schemas/blog` syntax.

sergioisidoro (approved):

sergioisidoro:
Merged! Thanks @codebender for taking the time to push this forward :)

PR #541, opened 2022-08-27
add spellchecking workflow to PRs

jtannas:
- add automatic spellchecking to PRs
- add a dictionary to deal with false-positives
- fix any actual issues
- use the default American spelling otherwise we'll keep having to update the dictionary with English equivalents (eg. behavior vs behaviour)


sergioisidoro (approved):
Looks good!

PR #542, opened 2022-08-28
drop support for EOL ruby + rails versions

jtannas:
I'm just working on getting my development environment going for this gem. I figure it'd be nice to not have to worry about Ruby versions older than what's in the `.ruby_version` file.

Hiccup 1: Looks like there's issue with RubyGems for Ruby v2.7.6 that causes the `mini_racer` installation to fail.
Solution: https://github.com/rubyjs/mini_racer/issues/220#issuecomment-1010724771

Hiccup 2: Ruby 2.7.6 failed to build on my machine running Ubuntu 22.04 due to Ubuntu dropping support for old SSL versions
Solution: Updating ruby-build via `git -C "$(rbenv root)"/plugins/ruby-build pull` solved it for me


sergioisidoro:
Should we also bump the rails version in the docker file? that is currently in `ENV RAILS_VERSION 7.0.1`

I'm up for it, since it's past its security updates. But having updated an environment out of Rails 5 a few weeks ago, I feel that there might be people still using rails 5.x for a few more months. Having the tests running (but allowing them to fail) could give those people some valuable info about updating / not updating this gem to the next versions of rswag. 

But given EOL of rails 5.x I'm ok with merging this like this as well 👍

epugh:
I love seeing the new activity and the upgrades...  I've been waiting for a chance to get my Rails app over to the OpenAPI world, and this project is key to that!

jtannas:
> Having the tests running (but allowing them to fail) could give those people some valuable info about updating / not updating this gem to the next versions of rswag.

Seems like a solid compromise - we're not going out of our way to break support for Rails 5 so it's pretty likely there'll still be relevant updates for them for awhile. I'll add it back into the test suite & gemfile while keeping the newer versions as the defaults.

jtannas:
Merging this now since it has some fixes needed to develop further on Ubuntu 22.04

PR #547, opened 2022-09-05
Update base path configuration for OpenAPI 3

### LINKED ISSUE #318: OAS v3 tests fail without `basePath`, but `basePath` generates invalid Swagger
With the removal of [`basePath`](https://swagger.io/docs/specification/api-host-and-base-path/) in OpenAPI v3, I now define API host(s) using the `servers` field, like this:

```ruby
RSpec.configure do |config|
  config.swagger_docs = {
    'v1/swagger.yaml' => {
      openapi: '3.0.1',
      info: {
        title: 'My API V1',
        version: 'v1'
      },
      servers: [
        {
          url: 'https://api.example.com/v1',
        },
      ],
      # ...
    },
  }
end
```

And I define my endpoints using rswag like this:

```ruby
describe 'My API', type: :request, swagger_doc: 'v1/swagger.yaml' do
  path '/resources' do
    # ...
  end
end
```

This results in the generation of the expected, valid Swagger docs.

However, attempting to run the integration tests using this configuration will fail with a routing error, because the version (previously the `basePath`) is no longer prefixed to the path when running the tests:

```ruby
     ActionController::RoutingError:
       No route matches [GET] "/resources"
```

(The expected generated path is `/v1/resources`.)

OTOH, including the `basePath` in the Swagger config, in addition to the `servers` key, results in passing tests; however, the generated Swagger docs fail validation, because they include the invalid `basePath` property:

```yaml
openapi: 3.0.1
info:
  title: My API V1
  version: v1
basePath: "/v1"
servers:
- url: https://api.example.com/v1
```

Any suggestions on how to resolve this?

romanblanco:
## Problem

API Host and base URL definition has changed from OAS2 to OAS3.

`schemes`, `host` and `basePath` definitions are being replaced by [Server Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#server-object)

This is causing issue https://github.com/rswag/rswag/issues/318.

Closes https://github.com/rswag/rswag/pull/397
Closes https://github.com/rswag/rswag/issues/318

## Solution

I'm building on top of @developingchris's approach from https://github.com/rswag/rswag/pull/397.
The solution is to parse the URL from the Server Object.

* [OAS2.0: API Server and Base URL](https://swagger.io/docs/specification/2-0/api-host-and-base-path/)
* [OAS3.0: API Server and Base URL](https://swagger.io/docs/specification/api-host-and-base-path/)

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
- https://github.com/rswag/rswag/issues/318

### Checklist
- [x] Added tests
- [x] Changelog updated
- [x] ~Added~ documentation to README.md

see https://github.com/rswag/rswag/issues/318

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -102,7 +102,15 @@ module Rswag
       end
 
       def add_path(request, metadata, swagger_doc, parameters, example)
-        template = (swagger_doc[:basePath] || '') + metadata[:path_item][:template]
+        template = ''
+        if doc_version(swagger_doc).start_with?('2')
+          template = (swagger_doc[:basePath] || '')
+        else
+          template = base_path_from_servers(swagger_doc)
+        end
+
+        template +=  metadata[:path_item][:template]
+
 
         request[:path] = template.tap do |path_template|
           parameters.select { |p| p[:in] == :path }.each do |p|
@@ -117,6 +125,14 @@ module Rswag
             path_template.concat(i.zero? ? '?' : '&')
             path_template.concat(build_query_string_part(p, example.send(p[:name]), swagger_doc))
           end
+        end
+      end
+
+      def base_path_from_servers(swagger_doc)
+        if swagger_doc[:servers].count == 1
+          URI(swagger_doc[:servers].first[:url]).path
+        else
+          URI(swagger_doc[:servers].first[:url]).path
         end
       end
 
@@ -150,7 +166,7 @@ module Rswag
                           when :spaceDelimited then '%20'
                           when :pipeDelimited then '|'
                           end
-              return "#{CGI.escape(name.to_s)}=" + value.to_a.flatten.map{|v| CGI.escape(v.to_s) }.join(separator) 
+              return "#{CGI.escape(name.to_s)}=" + value.to_a.flatten.map{|v| CGI.escape(v.to_s) }.join(separator)
             end
           else
             return "#{name}=#{value}"
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -101,16 +101,28 @@ module Rswag
         request[:verb] = metadata[:operation][:verb]
       end
 
+      def base_path_from_servers(swagger_doc, use_server = :default)
+        return '' if swagger_doc[:servers].nil? || swagger_doc[:servers].empty?
+        server = swagger_doc[:servers].first
+        variables = {}
+        server.fetch(:variables, {}).each_pair { |k,v| variables[k] = v[use_server] }
+        base_path = server[:url].gsub(/\{(.*?)\}/) { |name| variables[name.to_sym] }
+        URI(base_path).path
+      end
+
       def add_path(request, metadata, swagger_doc, parameters, example)
-        template = ''
-        if doc_version(swagger_doc).start_with?('2')
-          template = (swagger_doc[:basePath] || '')
-        else
-          template = base_path_from_servers(swagger_doc)
+        open_api_3_doc = doc_version(swagger_doc).start_with?('3')
+        uses_base_path = swagger_doc[:basePath].present?
+
+        if open_api_3_doc && uses_base_path
+          ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: basePath is replaced in OpenAPI3! Update your swagger_helper.rb')
         end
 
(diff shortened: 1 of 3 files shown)


romanblanco:
Hi @jtannas, could you look at the changes to see if it makes sense to you?

rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -101,8 +101,27 @@ def add_verb(request, metadata)
...
+        if doc_version(swagger_doc).start_with?('3')
+          if !swagger_doc[:basePath].nil?
+            ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: basePath is replaced in OpenAPI3! Update your swagger_helper.rb')
+            template = (swagger_doc[:basePath] || '') + metadata[:path_item][:template]
+          else
+            template = base_path_from_servers(swagger_doc) + metadata[:path_item][:template]
+          end
+        else # OpenAPI 2
+          template = (swagger_doc[:basePath] || '') + metadata[:path_item][:template]
+        end
```
jtannas: If not very good at understanding nested conditionals so I rearranged the code to help myself understand.
Your call on which version you'd prefer to add since you did the heavy lifting though :slightly_smiling_face: 

```suggestion
        is_swagger3 = doc_version(swagger_doc).start_with?('3')
        uses_swagger2 = !swagger_doc[:basePath].nil?
        base_path = if uses_swagger2
                      ActiveSupport::Deprecation.warn(<<~DEPRECATION) if is_swagger3
                        Rswag::Specs: WARNING: basePath is replaced in OpenAPI3! Update your swagger_helper.rb
                      DEPRECATION
                      (swagger_doc[:basePath] || '')
                    else
                      base_path_from_servers(swagger_doc)
                    end
        template = base_path + metadata[:path_item][:template]
```

rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -101,8 +101,27 @@ def add_verb(request, metadata)
...
       end
 
+      def base_path_from_servers(swagger_doc, use_server = :default)
+        return '' if swagger_doc[:servers].nil? || swagger_doc[:servers].empty?
+        base_paths = swagger_doc[:servers].map do |server|
+          variables = {}
+          server.fetch(:variables, {}).each_pair { |k,v| variables[k] = v[use_server] }
+          server[:url].gsub(/\{(.*?)\}/) { variables[$1.to_sym] }
+        end
+        URI(base_paths.first).path
```
jtannas: why map over all the servers if only the first one is being consumed? :think

rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -101,8 +101,27 @@ def add_verb(request, metadata)
         request[:verb] = metadata[:operation][:verb]
       end
 
+      def base_path_from_servers(swagger_doc, use_server = :default)
+        return '' if swagger_doc[:servers].nil? || swagger_doc[:servers].empty?
+        base_paths = swagger_doc[:servers].map do |server|
+          variables = {}
+          server.fetch(:variables, {}).each_pair { |k,v| variables[k] = v[use_server] }
+          server[:url].gsub(/\{(.*?)\}/) { variables[$1.to_sym] }
```
jtannas: I *think* this will work and is more readable but I haven't tested it
```suggestion
          server[:url].gsub(/\{(.*?)\}/) { |variable_name| variables[variable_name.to_sym] }
```

jtannas (commented):
Thanks for the contribution! :slightly_smiling_face: 
I really appreciate the help while I'm getting this all figured out.

Just a few notes on places I found difficult to follow and it should be good to go!

rswag-specs/lib/rswag/specs/request_factory.rb
romanblanco: You're right. This was a result of my confusion with supporting multiple servers and it is not necessary.

rswag-specs/lib/rswag/specs/request_factory.rb
romanblanco: I've tried to reduce the complexity a bit, see if it's better :wink:

romanblanco:
@jtannas Thanks for keeping the Gem alive!

I've applied the suggested changes ([diff](https://github.com/rswag/rswag/compare/5dbab7aca1a3eaabff1c291593f84c74991fa27b..2264b033e8ff7c88d04a800c9197f7dc8971ac47)). Let me know if you see more things to improve :slightly_smiling_face:

PR #550, opened 2022-09-09
Make the docker user non-root for easier volume sharing

jtannas:
The `swagger.json`/`Gemfile.lock` files generated inside of the test-app are created by the docker user. The docker user is root by default. When using docker-compose volumes to share files between the host & the container, the generated file retains the root docker user as the owner. This prevents any edits from being made to those files locally without using either `sudo` or `sudo chown`.

The solution is to set the container user id to match your own user id. Typically you'll be `USER_ID=1000` on your own machine, and you can double check that using `$ id -u`.
If your user id differs, then `docker build --build-arg USER_ID=$(id -u) .` will pass your id into the docker container build.


jtannas:
Draft pending merge of update-changelog-2.6.0 to master

BookOfGreg (approved):
Good use of docker best practice there.

PR #556, opened 2022-09-24
Add ADDITIONAL_RSPEC_OPTS ENV var for rswag:specs:swaggerize

asavageiv:
This allows easier filtering of rswag specs that are intermingled with non-rswag specs. For example, if run_test! is used consistently, then one can do the following to run only the rswag specs.

ADDITIONAL_RSPEC_OPTS='--tag rswag'
rake rswag:specs:swaggerize

## Problem
In my work repository, rswag specs are intermingled with non-rswag specs in the `requests` subfolder. The non-rswag specs are run along side the rswag specs with `rake rswag:specs:swaggerize` and slows down document generation quite a bit, running irrelevant tests.

## Solution
Allow passing additional rspec options via ENV variable so that only the rswag specs can be run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Add tooling for measuring test coverage so that changes are safer (https://github.com/rswag/rswag/pull/551)
 - Add CSP compatible with rswag in case the Rails one is not compatible (https://github.com/rswag/rswag/pull/263)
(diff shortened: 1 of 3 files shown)


jtannas:
I like the idea! Add a changelog note and I'd say it's good to go :slightly_smiling_face:

asavageiv:
> I like the idea! Add a changelog note and I'd say it's good to go 🙂

Done. I also added some usage instructions to the README.

jtannas (approved):

PR #571, opened 2022-11-06
update changelog with v2.7.0+v2.8.0 changes

jtannas:


romanblanco (commented):
The [Compatibility table](https://github.com/rswag/rswag/blob/master/README.md?plain=1#L21-L29) in README could also be updated

romanblanco:
and [SECURITY.md](https://github.com/rswag/rswag/blob/master/SECURITY.md?plain=1#L7)

jtannas:
> The [Compatibility table](https://github.com/rswag/rswag/blob/master/README.md?plain=1&rgh-link-date=2022-11-07T19%3A56%3A40Z#L21-L29) in README could also be updated

Not sure what needs doing here :thinking: It looks like the intent is to track when the supported swagger & swagger-ui versions change and I don't know if Swagger 3.1 support is complete yet. Suggestions?

romanblanco:
I think that https://github.com/rswag/rswag/issues/464 indicates that ~~Swagger~~ OpenAPI 3.1 is not supported yet and that there's still work to do.

~~I'd keep the current numbers (`3.0.3 | 3.52.5`) for the latest rswag versions.~~

EDIT:

swagger-ui 4.1.3 was merged in https://github.com/rswag/rswag/pull/476 and released in 2.6.0.

Actually, I'm thinking that the table could be removed as well. It's confusing, since the latest version should still be compatible with Swagger 2.0 :thinking:

CHANGELOG.md
```diff
@@ -8,6 +8,34 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+### Changed
+
+### Fixed
+
+### Documentation
+
+## [2.8.0]
```
romanblanco, line 17: I think that the changes for 2.8.0 should be added separately, after the 2.7.0 is released. 

It could be confusing to release version 2.7.0 with CHANGELOG for 2.8.0 :wink:

SECURITY.md
```diff
@@ -2,16 +2,8 @@
...
-At the current point in time this means 2.5.x
-
-<!--
-| Version | Supported          |
-| ------- | ------------------ |
-| 2.5.x   | :white_check_mark: |
-| < 2.5.0 | :x:                |
--->
+Only the latest version is supported due to the small size of the team.
+[Follow this link to see the latest release](https://github.com/rswag/rswag/releases/latest)
```
romanblanco, line 6: :+1:

CHANGELOG.md
jtannas: 2.8.0 has been tagged so the change list is complete there; merging & releasing 2.8.0

PR #572, opened 2022-11-06
:construction: Rubocop cleanup

jtannas:
Getting the code to a point where rubocop can be run automatically on new PRs. The goal is to prevent PR review time being spent on formatting.


jtannas:
This ended up being a lot more manual changes than intended. I'm going to restart and split it up to clearly show the different steps via atomic commits and by keeping manual changes for follow up PRs.

PR #574, opened 2022-11-08
cleanup: Drop support for OpenAPI v2

### LINKED ISSUE #609: Dropping of OAS2
### Discussed in https://github.com/rswag/rswag/discussions/503

<div type='discussions-op-text'>

<sup>Originally posted by **sergioisidoro** February 22, 2022</sup>
Hi everyone, 

Right now supporting OAS2 adds quite a big overhead, and I'm wondering if it really makes sense to continue supporting it. OAS 3.0.0 was released on 26 July 2017, and if we to continue to keep this gem in good shape, it's time to focus our resources into supporting OAS >= 3

Some examples of the overhead in keeping backwards compatibility include recent pull requests, like:
* Response examples - https://github.com/rswag/rswag/pull/501
* Query - https://github.com/rswag/rswag/pull/499 

What do you think? Please react with 👎 if still need OAS2 support.</div>

romanblanco:
## Problem
As mentioned in https://github.com/rswag/rswag/discussions/503, maintaining OpenAPI v2 specification is becoming too big overhead to support. 

Fixes https://github.com/rswag/rswag/issues/609

## Solution
This PR is intended to drop all code related to OpenAPI v2 and end its support.

Also, because the Swagger has been renamed to OpenAPI, the codebase and documentation is also being updated in order to avoid confusion in terminology. 

* [Swagger 2.0 (aka OpenAPI 2.0) ](https://swagger.io/specification/v2/)
* [OpenAPI 3.0.0](https://spec.openapis.org/oas/v3.0.0.html)

- [ ] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
- https://github.com/rswag/rswag/discussions/503

### Checklist
- [x] Added tests
- [x] Changelog updated


romanblanco:
Rejecting OpenAPI v2 specification should be added to the PR.

test-app/openapi/v1/openapi.json
```diff
@@ -14,7 +14,9 @@
         "operationId": "testBasicAuth",
         "security": [
           {
-            "basic_auth": []
+            "basic_auth": [
+
+            ]
```
jtannas, line 19: any idea why the three newlines have been added in this file to the empty arrays?

rswag-specs/spec/rswag/specs/request_factory_spec.rb
```diff
@@ -589,38 +557,12 @@ module Specs
...
         context 'base path' do
-          context 'openapi 2.0' do
-            before { swagger_doc[:basePath] = '/api' }
-
-            it 'prepends to the path' do
-              expect(request[:path]).to eq('/api/blogs')
-            end
-          end
-
           context 'openapi 3.0' do
```
jtannas: I'm sure sure this `context` block is still necessary if there's just the one test within `context 'base path'` 🤔 
Thoughts?

jtannas (commented):
Thanks for taking this on 🙇‍♂️ 
Lots of great progress here with some minor notes:
- There are a few places where `swagger` is still used instead of `openapi` (eg.` swagger_dry_run`). Is this intentional? If not it's probably worth doing a find/replace sweep of the project
- Might be worth starting a `MIGRATION_GUIDE.md` file for `2.x -> 3.0` given that we're renaming a lot and dropping support for OpenAPI 2 / Rspec 2 / Ruby 2.6

romanblanco:
> There are a few places where swagger is still used instead of openapi (eg. swagger_dry_run). Is this intentional? If not it's probably worth doing a find/replace sweep of the project

@jtannas I've tried to sweep through the repo and identify anything that made sense to rename or remove. There are still some mentions of Swagger in `README.md`, but I think they are still relevant for documentation purpose.

I also see updating `README.md` on the TODO list (moving content to Wiki) and would suggest to address the changes in a follow-up PR focusing on documentation.

I'll take a second look at `swagger_dry_run`, but for example `swaggerize` I've kept intentionally. Something like `openapize` did not feel right and I didn't feel like a change would bring any clarity or improvement to the code.

romanblanco:
I've just checked with `grep -rnl 'swagger'` and still can see some places to change (e.g. `rswag-api/spec/generators/rswag/api/tmp/config/initializers/rswag_api.rb`), so I'll update that soon.

jtannas (approved):

mattpolito:
This is a long running branch with a lot of changes... where are we at on this one? I'm guessing we'll have to do another pass since there are changes in master that appear to be removed here.

romanblanco:
> This is a long running branch with a lot of changes... where are we at on this one?

@mattpolito, besides responding to the review suggestions (renaming `swagger_dry_run`, starting `MIGRATION_GUIDE.md`,...), I think the changes here are not far from being done.

However, together with this, in 3.0.0 we also need to take care of updating [other deprecations](https://github.com/rswag/rswag/pull/574/files#diff-7721ef414798cc5cd867aaa891fe5e0805bd238d98e2f3535a51e88d68b0c4efR10-R16) (Ruby, Rspec). It would be better to have the other changes ready as well (see https://github.com/rswag/rswag/pull/625, https://github.com/rswag/rswag/pull/636), when we decide to merge this one.

I'm keeping the PR updated with latest merges in the meantime.

romanblanco:
@mattpolito @jtannas, I consider this change ready. 

I'm wondering about what should we put into the migration document.

The summary of this PR would probably be

```
- "Swagger" has been rebranded to "OpenAPI"
  - `SwaggerFormatter` is now `OpenapiFormatter`
  - `swagger_root` is now `openapi_root`
  - `swagger_strict_schema_validation` is now `openapi_strict_schema_validation`
  - `swagger_helper` is now `openapi_helper`
  - `swagger_endpoint` is now `openapi_endpoint`
  - `swagger_filter` is now `openapi_filter`

  - `swagger_dry_run` is now `rswag_dry_run`  # !!!
  - `swagger_doc` is now `openapi_spec`  # !!!
  - `swagger_docs` is now `openapi_specs`  # !!!

  - The default folder for generating OpenAPI spec has moved from `swagger` to `openapi`.
  - `swaggerize` task has kept it's name
```

Other changes (e.g. moving `#/definitions/` under `#/components/schemas/`) would be the migration of Swagger 2.0 to OpenAPI 3.0 itself.

Not sure we need to document that part here too :thinking:.

mattpolito:
Awesome I'll take a look this weekend to get a another comb through of it. If we're changing everything from swagger to openapi, I'd recommend us changing the task to something different as well and aliasing swaggerize to it.

romanblanco:
@mattpolito Thank you! I'll be off next week, but will try to get back to this soon after.

mattpolito (commented):
Have we considered the upgrade path for these changes? Dropping support for OpenAPI v2 is completely different than renaming so many things.  For example has there been deprecation warnings put in place to say that `swagger_root` & `swagger_endpoint` are going away? That's going to break every install.

I know a lot of work has been put into this but the support for v2 should be dropped without renaming public api methods.

Then deprecation warnings put in saying that those methods will be changed in the next release.

If it comes along with a major version bump we could maybe throw caution to the wind but might be a bit bad form.

rswag-specs/spec/rswag/specs/request_factory_spec.rb
romanblanco: True. Removing it, since now that only OpenAPI v3 is supported, it does not make sense to separate the context.

Thanks!

test-app/openapi/v1/openapi.json
romanblanco: Don't know really :confused: 
Didn't find anything I've removed in this PR that would cause this.

romanblanco:
> Have we considered the upgrade path for these changes? Dropping support for OpenAPI v2 is completely different than renaming so many things. For example has there been deprecation warnings put in place to say that `swagger_root` & `swagger_endpoint` are going away? That's going to break every install.
> 
> I know a lot of work has been put into this but the support for v2 should be dropped without renaming public api methods.
> 
> Then deprecation warnings put in saying that those methods will be changed in the next release.
> 
> If it comes along with a major version bump we could maybe throw caution to the wind but might be a bit bad form.

@mattpolito These changes would come along with a major version, but I agree. We can do better with letting folks know that we're renaming the methods in v3.

The PR is now aiming to be merged into v3 branch and I'll create a PR into current master, to warn about the change.

romanblanco:
@mattpolito, since these changes are going into `v3` branch, I'll go ahead and merge to unblock other PRs.

The PR adding warning about upcoming changes in `v3` is out: https://github.com/rswag/rswag/pull/688

test-app/openapi/v1/openapi.json
romanblanco: I see the same happened in this commit: https://github.com/rswag/rswag/pull/675/commits/d65f4fe4db17550809d462fd35d6edc2010dc6d2, so I'm suspecing some change in dependency that rswag uses has caused this.

PR #583, opened 2022-11-13
Allow `responses` global key alongside `definitions` for swagger 2.0

DoubleJarvis:
## Problem
Can't describe endpoint response schema using global `responses` section as a `$ref`.

## Solution
When fetching `'definitions'` from swagger_doc we should also fetch `'responses'`.

* [Describing Responses](https://swagger.io/docs/specification/2-0/describing-responses/). Specifically look for "Reusing Responses" section.

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

This change is only relevant to swagger 2.0.

### Related Issues
Links to any related issues.

### Checklist
- [x] Added tests
- [x] Changelog updated

Consider the following
swagger_helper.rb
```ruby
RSpec.configure do |config|
  config.swagger_root = '.'
  
  config.swagger_docs = {
    'swagger.json' => {
      "swagger": "2.0",
      "responses": {
        "NotFound": {
          "description": "Object not found"
        }
      }
    }
  }
end
```
then the following should work
spec/requests/whatever_spec.rb
```ruby
require 'swagger_helper'

RSpec.describe 'Whatever API', type: :request do
  path '/whatever' do
    get 'Gets Whatever' do
      response 404, 'Whatever not found' do
        schema '$ref': '#/responses/NotFound'

        run_test!
      end
    end
  end
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Make dependency on rspec-core explicit instead of implied (https://github.com/rswag/rswag/pull/554)
 - Fix base path for OAS3 specification (https://github.com/rswag/rswag/pull/547)
 - Fix ResponseValidator adding support for nullable and required headers (https://github.com/rswag/rswag/pull/527)
+- Fix inability to define common responses in the global responses section and reference that definition via $ref at the operation level in swagger 2.0
 
 ### Documentation
 
--- a/rswag-specs/lib/rswag/specs/response_validator.rb
+++ b/rswag-specs/lib/rswag/specs/response_validator.rb
@@ -72,7 +72,7 @@ module Rswag
 
       def definitions_or_component_schemas(swagger_doc, version)
         if version.start_with?('2')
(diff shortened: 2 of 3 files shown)


jtannas:
Hi @DoubleJarvis 
Sorry for the slow reply. We're actually [discussing dropping support for OAS2.0](https://github.com/rswag/rswag/discussions/503)  so I'm hesitant about adding more features to it only to pull the rug out from it not long after. Is there something missing in our OAS3 support preventing you from upgrading?
Cheers

DoubleJarvis:
> Sorry for the slow reply

No worries, mate.
 Unfortunately I haven't seen said discussion and PR before making this request. I'm using rswag at work, sadly it's not my call when we're going to upgrade to OAS3. I don't think there's anything preventing it, [we're just too busy dragging a box around instead of taking some time to install a wheel :) ](https://www.renaissance.ie/wp-content/uploads/2018/02/square-wheels-round-wheels.jpg)
I agree, in your transition period new PRs are just going to add unnecessary work (need to rebase and whatnot), so I'll close this one.
I'll use this deprecation as another reason to push for OAS3 upgrade.

PR #589, opened 2022-12-12
fix: mismatched types on compare

jiz4oh:
## Problem
`apiKey` type security is only valid on declare as Symbol, and is invalid as String

## Solution
convert type as symbol before compare

* [LINK TO RELEVANT OPEN API SPECS PAGE](https://spec.openapis.org/oas/v3.1.0#data-types)

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
Links to any related issues.

### Checklist
- [x] Added tests
- [x] Changelog updated

```rb
# spec/swagger_helper.rb
RSpec.configure do |config|
  config.swagger_root = Rails.root.to_s + '/swagger'

  config.swagger_docs = {
    'v1/swagger.json' => {
      components: {
        securitySchemes: {
          api_key: {
            type: 'apiKey',  # <= see here, declare as String
            name: 'api_key',
            in: :query
          }
        }
      }
    }
  }
end
```

```rb
# example of documenting an endpoint that handles basic auth and api key based security
describe 'Auth examples API' do
  path '/auth-tests/basic-and-api-key' do
    post 'Authenticates with basic auth and api key' do
      tags 'Auth Tests'
      operationId 'testBasicAndApiKey'
      security [{ basic_auth: [], api_key: [] }]

      response '204', 'Valid credentials' do
        let(:api_key) { 'foobar' }
        run_test!
      end

      response '401', 'Invalid credentials' do
        let(:api_key) { 'bar-foo' }
        run_test!
      end
    end
  end
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 - Fix ADDITIONAL_RSPEC_OPTS to always apply (https://github.com/rswag/rswag/pull/584)
+- Fix apiKey can not be declared as `String` (https://github.com/rswag/rswag/pull/589)
 
 ### Documentation
 
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -43,7 +43,7 @@ module Rswag
         schemes = security_version(scheme_names, swagger_doc)
 
         schemes.map do |scheme|
-          param = (scheme[:type] == :apiKey) ? scheme.slice(:name, :in) : { name: 'Authorization', in: :header }
+          param = (scheme[:type]&.to_sym == :apiKey) ? scheme.slice(:name, :in) : { name: 'Authorization', in: :header }
           param.merge(type: :string, required: requirements.one?)
         end
       end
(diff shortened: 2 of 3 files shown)


jiz4oh:
Hi there, could anyone suggest the next step? This fix help us to read "Global Metadata" from a file instead

romanblanco (commented):
@jiz4oh, Thank you for opening up a PR!

I'm trying to understand the motivation to change this.

Would you be able to describe more about how will this enable you to read the [global metadata](https://github.com/rswag/rswag#global-metadata) from file?

Besides, the PR would need tests that are failing if the change is taken back, properly updating [README.md](https://github.com/rswag/rswag#specifyingtesting-api-security) and other documentation and listing your PR in CHANGELOG.md.

jiz4oh:
Hi @romanblanco, thanks for your reply. 

I open up the PR because we set swagger_doc from an YAML file.
image

 The `RSpec.configuration.swagger_docs` is a hash which keys and values all are `String`, so I use `ActiveSupport::HashWithIndifferentAccess` to resolve the `String` key.  But it's not working with `String` values.
 
And the question is even through I declare some `apiKey`s in the YAML file, 
image 
I cannot add authorization in our rswag specs due to [here](https://github.com/rswag/rswag/blob/154af48db7f5b702b69091e4f1b25fb15c2e076c/rswag-specs/lib/rswag/specs/request_factory.rb#L46), it's required us declare `apiKey` as a `Symbol` instead of `String`


> Besides, the PR would need tests that are failing if the change is taken back

I have added some tests in my PR, could you give me more suggestions about failing tests?

> properly updating [README.md](https://github.com/rswag/rswag#specifyingtesting-api-security) and other documentation

There is a little bugfix, actually I have no idea how to describe this in document.

> and listing your PR in CHANGELOG.md.

Thanks, I will add it later

PR #591, opened 2023-01-05
Fix RequestFactory gsub for base_path_from_servers

saturnflyer:
## Problem
Commit 2264b033e8ff7c88d04a800c9197f7dc8971ac47 introduced a change to this method where the gsub replacement of the server url switched from the global Regexp match variable of $1 to the block argument for gsub.

This change made it so that the argument to the variable name lookup include the braces used in the string for the url variable.

That lead to the hash not returing any matching keys and values and the generated URI string of '://' being passed to the URI() method which raise an error.

## Solution

The solution is to ~~use `tr` to replace matching `{` and `}` characters from the block argument passed to `gsub!`~~ use the global Regex `$1` match variable in `Rswag::Specs::RequestFactory#base_path_from_servers`

* [Server Object specification](https://spec.openapis.org/oas/v3.1.0#server-object) does not list `protocol`

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
None

### Checklist
- [x] Added tests
- [x] Changelog updated

The configuration in the spec was updated to include a `protocol` value in the server variables with a `:default` key. Only this change is necessary to trigger an Invalid URI error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- Invalid URI error when specifying protocol within server configuration
 - Fix ADDITIONAL_RSPEC_OPTS to always apply (https://github.com/rswag/rswag/pull/584)
 
 ### Documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
-- Invalid URI error when specifying protocol within server configuration
+- Invalid URI error when specifying protocol within server configuration (https://github.com/rswag/rswag/pull/591)
 - Fix ADDITIONAL_RSPEC_OPTS to always apply (https://github.com/rswag/rswag/pull/584)
 
 ### Documentation
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -106,7 +106,7 @@ module Rswag
         server = swagger_doc[:servers].first
         variables = {}
         server.fetch(:variables, {}).each_pair { |k,v| variables[k] = v[use_server] }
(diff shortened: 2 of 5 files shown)


rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -106,7 +106,7 @@ def base_path_from_servers(swagger_doc, use_server = :default)
         server = swagger_doc[:servers].first
         variables = {}
         server.fetch(:variables, {}).each_pair { |k,v| variables[k] = v[use_server] }
-        base_path = server[:url].gsub(/\{(.*?)\}/) { |name| variables[name.to_sym] }
+        base_path = server[:url].gsub(/\{(.*?)\}/) { |name| variables[name.tr('{}','').to_sym] }
```
gridar: Why not use the `$1` result filled with the regex's group ?
```suggestion
        base_path = server[:url].gsub(/\{(.*?)\}/) { variables[$1.to_sym] }
```
it remove the need to use the `tr`

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- Invalid URI error when specifying protocol within server configuration
```
romanblanco:
```suggestion
- Invalid URI error when specifying protocol within server configuration (https://github.com/rswag/rswag/pull/591)
```

romanblanco (commented):
@saturnflyer Thank you for fixing it! 

Please, just update the `CHANGELOG.md` with the PR number and document the change in [`README.md`](https://github.com/rswag/rswag/blob/master/README.md?plain=1#L316) examples. Than it's good to be merged :+1:

romanblanco:
A change in [test-app](https://github.com/rswag/rswag/blob/master/test-app/spec/swagger_helper.rb#L93) would also be nice for demonstration.

romanblanco:
Fixes https://github.com/rswag/rswag/issues/597
Fixes https://github.com/rswag/rswag/issues/601

romanblanco (approved):
LGTM :+1:, Thank you, @saturnflyer!

romanblanco (approved):

PR #604, opened 2023-03-05
Strict schema validation

### LINKED ISSUE #163: Strict schema validation
Hello 👋 

First of all - thanks for the awesome gem!

I'd like to know your thoughts about adding support for a schema validation. Per json-schema gem passing `strict: true` option will make sure we won't allow undefined properties to exist in our schemas.

    schema = {
      "type" => "object",
      "required" => ["a"],
      "properties" => {
        "a" => { "type" => "integer" }
      }
    }

    JSON::Validator.validate(schema, { "a" => 1 }, strict: true)
    #=> true
    JSON::Validator.validate(schema, { "a" => 1, "b" => 2 }, strict: true)
    #=> false

Not sure about best API for that feature though. We can allow for a global setting, but with an option to override this per individual specs too. Something like the following maybe?

    run_test!(strict: true)

IlkhamGaysin:
## Problem
I see some problems in ruby projects - it's data bloating. As projects grow they get complicated so it's hard to keep responses clean. It happed to projects I am working on and others(see the issue). So I decided to bring the changes back from initial issue.

## Solution
I want to add a new configuration option on top of the main called `additionalProperties` from [json-schema](https://github.com/voxpupuli/json-schema) to allow users to validate their schemas strictly.

* [EXAMPLE_LINK TO RELEVANT OPEN API SPECS PAGE](https://spec.openapis.org/oas/v3.1.0#data-types)

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues

Closes https://github.com/rswag/rswag/issues/163

### Checklist
- [x] Added tests
- [x] Changelog updated

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -115,6 +115,12 @@ module Rswag
         )
       end
 
+      #
+      # Perform request and assert response matches swagger definitions
+      #
+      # @param strict: nil [Boolean] whether to validate response against given schema strictly
+      # @param &block [Proc] you can make additional assertions within that block
+      # @return [void]
       def run_test!(&block)
         if RSPEC_VERSION < 3
           ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: Support for RSpec 2.X will be dropped in v3.0')
@@ -123,7 +129,7 @@ module Rswag
           end
 
           it "returns a #{metadata[:response][:code]} response", rswag: true do
-            assert_response_matches_metadata(metadata)
+            assert_response_matches_metadata(metadata.merge(options))
             block.call(response) if block_given?
           end
         else
@@ -132,7 +138,7 @@ module Rswag
           end
 
           it "returns a #{metadata[:response][:code]} response", rswag: true do |example|
-            assert_response_matches_metadata(example.metadata, &block)
+            assert_response_matches_metadata(example.metadata.merge(options), &block)
             example.instance_exec(response, &block) if block_given?
           end
         end
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -118,10 +118,10 @@ module Rswag
       #
       # Perform request and assert response matches swagger definitions
       #
-      # @param strict: nil [Boolean] whether to validate response against given schema strictly
+      # @param swagger_strict_schema_validation: nil [Boolean] whether to validate response against given schema strictly
       # @param &block [Proc] you can make additional assertions within that block
       # @return [void]
-      def run_test!(&block)
+      def run_test!(**options, &block)
         if RSPEC_VERSION < 3
           ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: Support for RSpec 2.X will be dropped in v3.0')
           before do
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -122,14 +122,16 @@ module Rswag
       # @param &block [Proc] you can make additional assertions within that block
       # @return [void]
       def run_test!(**options, &block)
+        options[:rswag] = true unless options.key?(:rswag)
+
         if RSPEC_VERSION < 3
           ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: Support for RSpec 2.X will be dropped in v3.0')
           before do
             submit_request(example.metadata)
           end
 
-          it "returns a #{metadata[:response][:code]} response", rswag: true do
-            assert_response_matches_metadata(metadata.merge(options))
+          it "returns a #{metadata[:response][:code]} response", **options do
+            assert_response_matches_metadata(metadata)
             block.call(response) if block_given?
           end
         else
@@ -137,8 +139,8 @@ module Rswag
             submit_request(example.metadata)
           end
 
-          it "returns a #{metadata[:response][:code]} response", rswag: true do |example|
-            assert_response_matches_metadata(example.metadata.merge(options), &block)
+          it "returns a #{metadata[:response][:code]} response", **options do |example|
+            assert_response_matches_metadata(example.metadata, &block)
             example.instance_exec(response, &block) if block_given?
           end
         end
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -56,16 +56,16 @@ module Rswag
       end
 
 
-      def request_body_example(value:, summary: nil, name: nil) 
-        if metadata.key?(:operation) 
+      def request_body_example(value:, summary: nil, name: nil)
+        if metadata.key?(:operation)
           metadata[:operation][:request_examples] ||= []
-          example = { value: value } 
-          example[:summary] = summary if summary 
+          example = { value: value }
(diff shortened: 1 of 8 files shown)


IlkhamGaysin:
@jtannas 
Could you please take a look, thank you so much! 🍻

frostmark (approved):

IlkhamGaysin:
@romanblanco I've fixed conflicts and updated the branch

rusllonrails:
Looks good :+1: Plus 1 for merging :1st_place_medal:

romanblanco:
[GitHub had issues](https://www.githubstatus.com/incidents/52z0j6phhnjs) earlier today. Re-opening the PR to try and run CI tests.

romanblanco (changes requested):
@IlkhamGaysin, thank you for the PR!

I like the suggestion, but I've tried to run the tests locally and they're failing as well as in the CI.

Please, take a look at https://github.com/rswag/rswag/blob/master/CONTRIBUTING.md#test and update the PR.

romanblanco:
I'm also thinking that it would be really nice if you could update the [`test-app`](https://github.com/rswag/rswag/tree/e36229292ce0533a59ce9b119b197a2d2a2ed21c/test-app) to provide an example of use of the `strict` option.

IlkhamGaysin:
@romanblanco I will fix the specs and add a test case, thank you. I had a problem with docker locally so I relied on CI...

IlkhamGaysin:
@romanblanco I've fixed the specs and added the test case. The GitHub Actions has issues but locally the test suites aregreen

rswag-specs/lib/rswag/specs/example_group_helpers.rb
```diff
@@ -132,7 +138,7 @@ def run_test!(&block)
           end
 
           it "returns a #{metadata[:response][:code]} response", rswag: true do |example|
-            assert_response_matches_metadata(example.metadata, &block)
+            assert_response_matches_metadata(example.metadata.merge(options), &block)
```
a-lavis: Instead of merging `**options` with `metadata`, could we pass `**options` to `it`?

```ruby
      def run_test!(**options, &block)
        options[:rswag] = true unless options.key?(:rswag)

        if RSPEC_VERSION < 3
          ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: Support for RSpec 2.X will be dropped in v3.0')
          before do
            submit_request(example.metadata)
          end

          it "returns a #{metadata[:response][:code]} response", **options do
            assert_response_matches_metadata(metadata)
            block.call(response) if block_given?
          end
        else
          before do |example|
            submit_request(example.metadata)
          end

          it "returns a #{metadata[:response][:code]} response", **options do |example|
            assert_response_matches_metadata(example.metadata, &block)
            example.instance_exec(response, &block) if block_given?
          end
        end
      end
```

This should work, according to the README:
https://github.com/rswag/rswag/blob/4043166c6b3d691e3027df61fad3a5638b7256b5/README.md?plain=1#L276-L278

This would allow us to pass in other options to `it`, which is a feature I'd like to have.  For example, this would allow us to do the following:
```ruby
run_test! focus: true
```

I could put this up in a separate PR after this one is merged if folks would prefer that, but I thought I'd mention it here since we're already adding the `**options` parameter.

IlkhamGaysin: I really like the idea 👍

a-lavis: Thank you!  I put up a sub-PR here if that is helpful:
https://github.com/IlkhamGaysin/rswag/pull/1

IlkhamGaysin: @romanblanco I'm gonna merge changes of @a-lavis to my PR to push it all together 
are you ok with that?

romanblanco: @IlkhamGaysin, sounds good!

romanblanco: @IlkhamGaysin, I'll need you to rebase your branch once more because of the recent merge.

rswag-specs/lib/rswag/specs/example_group_helpers.rb
```diff
@@ -115,14 +115,22 @@ def example(mime, name, value, summary=nil, description=nil)
         )
       end
 
-      def run_test!(&block)
+      #
+      # Perform request and assert response matches swagger definitions
+      #
+      # @param swagger_strict_schema_validation: nil [Boolean] whether to validate response against given schema strictly
```
IlkhamGaysin: TODO: correct

rswag-specs/lib/rswag/specs/example_group_helpers.rb
IlkhamGaysin: @romanblanco done!

romanblanco (approved):
LGTM :+1: 

@IlkhamGaysin, @a-lavis, thank you guys!

romanblanco:
Hey @IlkhamGaysin,

I'm trying to use this feature in a new API spec I'm working on, and I'm running into a problem with fields that are not required.

In `README.md` we have an example of how to mark required fields:
https://github.com/rswag/rswag/blob/0edbe25fe61de72935303cdeb0a0447664492f9d/README.md?plain=1#L117

Using the same approach to describe pagination schema:
```
LINKS = {
  type: 'object',
  properties: {
    first: {
      type: 'string'
    },
    last: {
      type: 'string'
    },
    previous: {
      type: 'string'
    },
    next: {
      type: 'string'
    },
  },
  required: ['first', 'last']
}.freeze
```

If I have an example request that only has one page, I'm getting back error message:
```
...
       The property '#/links' did not contain a required property of 'previous' in schema 3efaf5d2-6a88-5939-8604-d152fa0e4574#
       The property '#/links' did not contain a required property of 'next' in schema 3efaf5d2-6a88-5939-8604-d152fa0e4574#
...
```

I suspect, that [parametrizing](
https://github.com/rswag/rswag/pull/604/files#diff-e8fe18e2685dfaaa26eae084b0a027f021d3e9d560c338c8f4af8979b023e440R67) `JSON::Validator.fully_validate` with `{ strict: true }` does not take the `required` field into consideration.

I'm wondering if I'm using it wrong or if it is a bug. Any ideas?

a-lavis:
@romanblanco 

The `strict` option for the `json-schema` gem does two things:
1. All Properties are Required
2. Additional Properties are Not Permitted

Because of number 1, it will ignore what you set in the `required` field.  All fields are required regardless.

In the future, we should take advantage of this new feature in `json-schema`: https://github.com/voxpupuli/json-schema/pull/494

That will allow us to _either_ make All Properties Required _or_ make Additional Properties Not Permitted.

mattpolito:
We ran into that as well when we manually made json-schema strict. The findings were the same as @a-lavis. Strict schema validation makes all declared keys required regardless of what is defined as 'required' in rswag.

IlkhamGaysin:
@mattpolito @romanblanco Hi guys, sorry for the delay, I've been on vacation. The initial idea of the strict option is to strictly ask for the presence of all options listed. I can apply the new flexible changes from json-schema if you would like to have them.
To do that I need:
- update the dependency in the rswag
- add a new option and replace the existing one to fit the json-schema options

Are you ok with me doing that?

IlkhamGaysin:
Here is the first PR guys
https://github.com/rswag/rswag/pull/675
@romanblanco @mattpolito

IlkhamGaysin:
@mattpolito @romanblanco hey guys, here is the fix that will make the restrictions more flexible.
https://github.com/rswag/rswag/pull/721

PR #619, opened 2023-04-03
fix: add missing link to Content Security Policy

romanblanco:
Validation indicator has problems loading (see https://github.com/rswag/rswag/issues/174#issuecomment-1315774714)

## Problem
Link to [validator.swagger.io](https://validator.swagger.io/validator?url=https) is missing in Content Security Policy for UI middleware (merged in https://github.com/rswag/rswag/pull/263)

## Solution
Add link into CSP.

--

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
- https://github.com/rswag/rswag/issues/174
- https://github.com/rswag/rswag/pull/263

### Checklist
- [ ] Added tests
- [x] Changelog updated

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Invalid URI error when specifying protocol within server configuration (https://github.com/rswag/rswag/pull/591)
 - Fix ADDITIONAL_RSPEC_OPTS to always apply (https://github.com/rswag/rswag/pull/584)
+- Add missing link to Content Security Policy (https://github.com/rswag/rswag/pull/619)
 
 ### Documentation
 
--- a/rswag-ui/lib/rswag/ui/middleware.rb
+++ b/rswag-ui/lib/rswag/ui/middleware.rb
@@ -41,9 +41,9 @@ module Rswag
       end
 
       def csp
-        <<~POLICY.gsub "\n", ' '
+        <<~POLICY.tr "\n", ' '
           default-src 'self';
-          img-src 'self' data:;
+          img-src 'self' data: https://validator.swagger.io;
           font-src 'self' https://fonts.gstatic.com;
           style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;
           script-src 'self' 'unsafe-inline';


romanblanco:
@keniiig21, can you help verify this fixes [your issue](https://github.com/rswag/rswag/issues/174#issuecomment-1315774714)?

ndbroadbent:
Thanks @romanblanco, I was just about to add this and open a PR. I have an integration test that loads the swagger UI page and my tests were failing with a CSP error. This looks like it will fix the issue for me. Thanks!

ndbroadbent:
This CSP is also causing me a lot of other problems, since the page won't be able to connect to my API endpoints. Could we add a way to configure it and add our own domains?

richardonrails:
Same issue here... the "Try it out" functionality doesn't work at all for me due to CSP issues. It's supposed to just connect to my `api.` subdomain.

I get both of these issues:
> Refused to load the image 'https://validator.swagger.io/validator?url=...' because it violates the following Content Security Policy directive: "img-src 'self' data:".

> Refused to connect to 'https://api....' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.

richardonrails:
In case others are waiting on a full fix, here's how I got everything working for my Rails app, I added the following to workaround the issue:

```rb
# config/initializers/rswag_ui_csp_monkeypatch.rb
# Monkeypatch https://github.com/rswag/rswag/blob/master/rswag-ui/lib/rswag/ui/middleware.rb
# due to issues described in https://github.com/rswag/rswag/pull/619/files
module Rswag
  module Ui
    class Middleware
      # ORIGINAL:
      # def csp
      #   <<~POLICY.gsub "\n", ' '
      #     default-src 'self';
      #     img-src 'self' data:;
      #     font-src 'self' https://fonts.gstatic.com;
      #     style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;
      #     script-src 'self' 'unsafe-inline';
      #   POLICY
      # end

      def csp
        <<~POLICY.gsub "\n", ' '
          default-src 'self';
          img-src 'self' data: https://validator.swagger.io;
          connect-src 'self' https://api.MYSITE.com https://api.MYSITE.test;
        POLICY
      end
    end
  end
end
```

gobijan:
Just wanted to create a similar PR.
Note that it's more performant to use tr instead of gsub:

  ```
# Monkey Patch CSP to allow Swagger UI to load external fonts and images
  module Rswag
    module Ui
      class Middleware
        private
          def csp
            <<~POLICY.tr "\n", ' '
            default-src 'self';
            img-src 'self' data: https://validator.swagger.io;
            connect-src 'self' https://api.MYSITE.com https://api.MYSITE.test;
            POLICY
          end
      end
    end
  end
```

romanblanco:
> This CSP is also causing me a lot of other problems, since the page won't be able to connect to my API endpoints. Could we add a way to configure it and add our own domains?

@ndbroadbent, would you create a separate PR to address the configuration of `connect-src`, or create new issue to track this, please.

PR #622, opened 2023-04-08
Add custom spec description option to `run_test!`

mattpolito:
## Problem
Specifications generated by `run_test!` are essentially a hard coded string of "returns a RESPONSE_CODE response". If you need to customize this you have to break down into the individual parts that `run_test!` constructs for us.

```ruby
it "custom specification description here" do |example|
  submit_request(example.metadata)
  assert_response_matches_metadata(example.metadata)
end
```

## Solution
To facilitate configuration without the need to break out of the nicety that `run_test!` provides, we can now pass a `description` as the first argument. 

`run_test!("custom specification description here")`

A huge benefit to this is that when paired with auto generation of request/response examples, many example scenarios can be provided for documentation within the same `response` grouping.

NOTE: `description` is not an allowable option to pass down to a specification, so we delete it vs forwarding down to `it`.

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Checklist
- [x] Added tests
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added option --spec_path to the generator command with requests as default value (https://github.com/rswag/rswag/pull/607)
 - Add support for `:getter` parameter option to explicitly define custom parameter getter method and avoid RSpec conflicts with `include` matcher and `status` method (https://github.com/rswag/rswag/pull/605)
 - Added support strict schema validation and allow to pass metadata to run_test! (https://github.com/rswag/rswag/pull/604)
+- Add support for passing a custom specification description to `run_test!` (https://github.com/rswag/rswag/pull/622)
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -203,6 +203,14 @@ response '201', 'blog created' do
 end
 ```
 
+If you want to customize the description of the generated specification, the `description` option can be passed to **run_test!**
+
+```ruby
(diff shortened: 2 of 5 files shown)


mattpolito:
@romanblanco If rswag is in need of maintainer help. We can be of assistance.

a-lavis:
This is a great idea!  I have a proposal of how we could do it slightly differently here: https://github.com/hashrocket/rswag/pull/1
Let me know your thoughts!

mattpolito:
Updated PR with test-app scenario

BookOfGreg:
Hi @mattpolito, Roman pointed out your PR here on this project and I've chosen to grant you Write access to the project.
We hope you and the rest of the team at Hashrocket a good time with rswag.

mattpolito:
@romanblanco now that you cut a release for 2.9.0, this is technically a breaking change. I was hoping it would make it in prior to the 'run_test!' options change. How would you like to handle this?

romanblanco:
@mattpolito, my plan is to review and release this and other PRs in 2.10.0 before merging https://github.com/rswag/rswag/pull/574 and other future changes for 3.0.0.

I've tried to rebase your changes and run tests and everything is still passing.

mattpolito:
@romanblanco It's not a breaking change in code but in usage. However after thinking about how ruby would handle the args vs options, I believe it will be fine after-all.

romanblanco (approved):
LGTM :+1: 

@mattpolito, @a-lavis, @ayunker, thanks for getting this done!

PR #627, opened 2023-04-17
docs: add IRC channel for #rswag into README

romanblanco:
## Solution
Proposing IRC as a way to communicate more efficiently.

### Related Issues
https://github.com/rswag/rswag/discussions/495#discussioncomment-4080562

### Checklist
- ~Added tests~
- [ ] Changelog updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@ And that's not all ...
 
 Once you have an API that can describe itself in Swagger, you've opened the treasure chest of Swagger-based tools including a client generator that can be targeted to a wide range of popular platforms. See [swagger-codegen](https://github.com/swagger-api/swagger-codegen) for more details.
 
+## Communication
+
+Chat with us on [Libera.Chat](https://web.libera.chat/gamja/?channels=#rswag).
+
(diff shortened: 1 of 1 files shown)


mattpolito:
I'm open to give it a try but I wouldn't call IRC 'more efficient'.

mattpolito:
Unless I'm completely missing something... history is lost when I leave IRC. So I can't see how this would be a viable option to communicate for an open-source project. There needs to be a history for someone to catch up on. 

Please let me know if I'm missing something.

romanblanco:
@mattpolito, to get the history, one of the options is to use another platform (e.g. [Matrix](matrix.org)) that has a bridge to IRC.

I would've proposed using Matrix directly in the first place, but thought that using IRC is more inclusive and easier as one does not need to register an account to join the chat.

mattpolito:
@romanblanco Thanks for that info. I would say that if you need to use a specific client or platform to get history, then IRC would not be a helpful option. What were the motivating factors to moving toward a chat client?

romanblanco:
@mattpolito see https://github.com/rswag/rswag/discussions/495#discussioncomment-3609613

PR #628, opened 2023-04-19
Pass Metadata to HTTP Verb methods

a-lavis:
## Problem
Metadata can't be passed into HTTP verb methods.

## Solution
Add a keyword arguments parameter to the HTTP verb methods that is passed through to `describe`.

* [EXAMPLE_LINK TO RELEVANT OPEN API SPECS PAGE](https://spec.openapis.org/oas/v3.1.0#data-types)

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
n/a

### Checklist
- [x] Added tests
- [x] Changelog updated

Try passing keyword arguments to an HTTP verb method.

### DIFF
--- a/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
+++ b/rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
@@ -26,12 +26,24 @@ module Rswag
       end
 
       describe '#get|post|patch|put|delete|head|options|trace(verb, summary)' do
-        before { subject.post('Creates a blog') }
+        context 'when called without keyword arguments' do
+          before { subject.post('Creates a blog') }
 
-        it "delegates to 'describe' with 'operation' metadata" do
-          expect(subject).to have_received(:describe).with(
-            :post, operation: { verb: :post, summary: 'Creates a blog' }
-          )
+          it "delegates to 'describe' with 'operation' metadata" do
+            expect(subject).to have_received(:describe).with(
+              :post, operation: { verb: :post, summary: 'Creates a blog' }
+            )
+          end
+        end
+
+        context 'when called with keyword arguments' do
+          before { subject.post('Creates a blog', foo: 'bar') }
+
+          it "delegates to 'describe' with 'operation' metadata and provided metadata" do
+            expect(subject).to have_received(:describe).with(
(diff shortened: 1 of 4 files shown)


rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
```diff
@@ -159,13 +171,13 @@ module Specs
...
 
-          it "assigns all examples to the metadata" do 
-            expect(api_metadata[:operation][:request_examples].length()).to eq(2) 
+          it "assigns all examples to the metadata" do
+            expect(api_metadata[:operation][:request_examples].length()).to eq(2)
             expect(api_metadata[:operation][:request_examples][0]).to eq({ value: example_one, name: 0 })
             expect(api_metadata[:operation][:request_examples][1]).to eq({ value: example_two, name: example_two_name, summary: example_two_summary })
-          end 
-        end  
-      end
```
a-lavis, line 168: VSCode automatically deleted extra whitespace on lines 139-168.  This change is unrelated to the PR.  Let me know if you'd prefer me to not include these changes.

a-lavis:
This is a small enough change that I don't think it is necessary to update the README.  It isn't explicitly documented which methods support passing in metadata, but there are examples where this functionality is used, for example:
https://github.com/rswag/rswag/blob/5d77db93cecc2c2cabd93207049da67ae31756be/README.md?plain=1#L792

If folks think I need to update the README, maybe we could add a section to the README where we generally explain passing in metadata and list out the different methods that support it?

romanblanco:
@a-lavis Thank you for the PR, the changes look good!

> I don't think it is necessary to update the README. It isn't explicitly documented which methods support passing in metadata, but there are examples where this functionality is used

Since this is the case, I agree the `README.md` does not need to be updated.

Would you be able to provide an example of using this change in [test-app](https://github.com/rswag/rswag/tree/master/test-app) tough?

romanblanco (approved):

a-lavis:
@romanblanco Added an example to test-app in this commit: 9c6d88597a8e4ea6ac073905946ce39e37b9ea41

Let me know your thoughts!  I wasn't sure of the best place to shoehorn it in.

romanblanco (approved):
@a-lavis It's perfect!

Issue #402, opened 2021-02-16
Validation for unknown properties is missed


VitaliiLazebnyi:
Spec says:
```
When the test is executed, rswag also validates the actual response code and, where applicable, the response body against the provided JSON Schema.
```

My code:
```
      # schema
      components: {
        schemas: {
          endpoint: {
            type: :object,
            properties: {
              property1: { type: :boolean },
              property2: { type: :string }
            },
            required: %w[property1],
            "dependencies": {
              "property2": ["property1"]
            }
          }
        }
      }

      # spec
      response(200, 'existing records') do
        schema '$ref' => '#/components/schemas/endpoint'

        let(:id) { '1' }

        run_test! 
      end
```

When I pass into response `property3` the test still passes. 👎 
When I remove `property1` - the test fails 👍 
when I pass `property2` without `property1` - the test fails 👍 

So my question is: should I make any additional configuration to make the test fail if passed `property3` (because `property3` absent in schema). Or is it a bug that unknown properties passed without errors?

DominikAlberski:
I maybe wrong here but haw do you validate this in your controller? If controller accepts unknown properties then the test should pass. And from my understanding this schemas in test are just used to generate proper swagger documentation.

VitaliiLazebnyi:
```
I maybe wrong here but haw do you validate this in your controller? If controller accepts unknown properties then the test should pass. And from my understanding this schemas in test are just used to generate proper swagger documentation.
```

Please check [the documentation ](https://github.com/rswag/rswag)?
>>>When the test is executed, rswag also validates the actual response code and, where applicable, **the response body against the provided JSON Schema**.

I don't really understand what is the difference what exactly are my validations in controllers. A schema should validate response automatically according to the doc.
And it does. Except for cases when a service response contains unknown properties. (which are not included in a schema)

DominikAlberski:
Ok I misunderstood you there.

DominikAlberski:
I think I found what a problem could be.  You could try to set in your schema definition `"additionalProperties": false`

VitaliiLazebnyi:
@DominikAlberski 
you are right. It works with `"additionalProperties": false`
I was sure that all checks are done by `rswag` itself. but it looks like I was wrong.

jtannas:
Looks like this issue is resolved so I'm going to close it, but feel free to open another if you'd like any further help. Cheers 🙂

PR #644, opened 2023-06-01
feature: global setting to set additionalProperties as false

MichaelAlexanderBanuelos:
## Problem
Currently, Rswag does not provide a global setting to control the `additionalProperties` behavior in OpenAPI schemas. This means that `additionalProperties` is set to `true` by default, allowing any additional properties to be added to objects without validation. This can lead to unexpected data inconsistencies and potential security vulnerabilities. 

This provides a solution for issues like #402

## Solution
This pull request introduces a new global setting `disallow_additional_properties` in Rswag that allows users to set `additionalProperties: false` on all schemas by default, unless they have already been defined. This setting provides better control and enforces stricter validation on object properties throughout the API documentation.

Alternatively if we decide to not use this change or something similar, folks can still manually set `additionalProperties: false` but when working in a larger codebase its easy to forget these things.

### This concerns these parts of the OpenAPI Specification:
* [Data Types](https://spec.openapis.org/oas/v3.1.0#data-types)
* [Schema Object](https://spec.openapis.org/oas/v3.1.0#schema)

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
N/A

### Checklist
- [x] Added tests
- [x] Changelog updated

1. Configure the new global setting for `disallow_additional_properties: true`.
3. Run the swagger tests and verify that responses with undocumented keys fail the test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Allow passing metadata to HTTP verb methods (https://github.com/rswag/rswag/pull/628)
 - Added configuration for RuboCop RSpec to improve detection of RSpec examples and example groups (https://github.com/rswag/rswag/pull/632)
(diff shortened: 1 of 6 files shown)


pboling:
Is this really what is best for this project?

PR #652, opened 2023-06-27
Don't call outer `run_test!` API requests inside nested context

AlexMooney:
## Problem
When using a nested RSpec `context`, the outer `run_test!` cause unexpected API calls.

## Solution
Avoid touching the `before` during `run_test!` because that causes the API calls to be inherited.

### Related Issues
I searched for issues in GitHub but was not able to find any mention of this problem.

### Checklist
- [x] Added tests
- [x] Changelog updated
  - Not sure if I should do this as the current readme makes no mention of `context`
  - It's a fix for testing, so the test is itself an example of using it.

Run the changed specs without the change to example_group_helpers.rb and you'll have a `.count` of 2: one for line 23 and a second for line 28.

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -139,11 +139,8 @@ module Rswag
             block.call(response) if block_given?
           end
         else
-          before do |example|
-            submit_request(example.metadata)
-          end
-
           it description, *args, **options do |example|
+            submit_request(example.metadata)
             assert_response_matches_metadata(example.metadata, &block)
             example.instance_exec(response, &block) if block_given?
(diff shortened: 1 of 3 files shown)


lhmzhou (approved):

rswag-specs/lib/rswag/specs/example_group_helpers.rb
```diff
@@ -139,11 +139,8 @@ def run_test!(description = nil, *args, **options, &block)
             block.call(response) if block_given?
           end
         else
-          before do |example|
-            submit_request(example.metadata)
-          end
-
```
BrianHawley, line 145: Don't forget to make the same change to the `if RSPEC_VERSION < 3` code above.

Issue #653, opened 2023-06-29
[SECURITY] Path traversal exposing arbitrary yml and json files in the application


carlobeltrame:
I am reporting this security issue here as a public issue, as instructed in the SECURITY.md of this repo.
@BookOfGreg @oblakeerickson @romanblanco @sergioisidoro

## Describe the bug
On the latest version of rswag-api, the middleware [reads the path](https://github.com/rswag/rswag/blob/master/rswag-api/lib/rswag/api/middleware.rb#L15) of the requested URL, and if it is a .yml, .yaml or .json file [resolves this path against the swagger root](https://github.com/rswag/rswag/blob/master/rswag-api/lib/rswag/api/middleware.rb#L16) and [returns its contents](https://github.com/rswag/rswag/blob/master/rswag-api/lib/rswag/api/middleware.rb#L19..L28). Using a simple dot-dot-slash technique (and manually sending the http request instead of in a browser), this allows an attacker to read out any YAML or JSON files anywhere on the server. If rswag is used within a rails application, this includes config settings and secrets files.

## Steps to Test or Reproduce
I will not provide step-by-step reproduction instructions unless explicitly prompted to do so by the maintainers of this gem.

## Expected behavior
rswag-api should only ever expose the specific OpenAPI spec file(s) of the project. At the very least, the middleware should validate that the resolved path lies within the gem's directory. Better yet, it should only allow fetching explicitly allowlisted files. We already have to tell rswag-ui the name of the OpenAPI spec (using `.swagger_endpoint` inside a `Rswag::Ui.configure` initializer). The same information should also be used as an allowlist for rswag-api.

## Additional context
You should probably fix your issue templates. When trying to report a security vulnerability, the issue template selector just leads me to your SECURITY.md, which states to just open an issue.

## Dependency versions
The version of are you using for:
* Rswag: 2.8.0, but the bug is also present in the latest version 2.9.0
* RSpec: 3.10.0
* Rails: 6.1.7.1
* Ruby: 2.7

## Relates to which version of OAS (OpenAPI Specification)
- [ ] OAS2
- [ ] OAS3
- [ ] OAS3.1

sergioisidoro:
Thank you so much for taking the time to investigate and reporting this. 

Because rswag ui and rswag api are different gems, I don't think we can easily proceed with the proposed solution without a breaking change. I propose the path sanitisation, and trying to get a patch release to get this fixed asap to as many people as possible.

[...]

PR #654, opened 2023-06-30
Sanitize directory traversal in middleware

sergioisidoro:
Working on a mitigation for #653

### DIFF
--- a/rswag-api/lib/rswag/api/middleware.rb
+++ b/rswag-api/lib/rswag/api/middleware.rb
@@ -13,7 +13,11 @@ module Rswag
 
       def call(env)
         path = env['PATH_INFO']
-        filename = "#{@config.resolve_swagger_root(env)}/#{path}"
+        # Sanitize the filename for directory traversal by expanding, and matching the swagger root directory
+        filename = File.expand_path(File.join(@config.resolve_swagger_root(env), path))
+        unless filename.match Regexp.new('^' + Regexp.escape(@config.resolve_swagger_root(env)))
+          return @app.call(env)
+        end
 
         if env['REQUEST_METHOD'] == 'GET' && File.file?(filename)
           swagger = parse_file(filename)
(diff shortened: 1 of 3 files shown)


sergioisidoro:
I would really like some feedback on this, but given the urgency, I might self merge soon in case no one is able to provide a review. I would then try to make a patch release. @jtannas @romanblanco

jtannas (approved):

rswag-api/lib/rswag/api/middleware.rb
```diff
@@ -13,7 +13,11 @@ def initialize(app, config)
 
       def call(env)
         path = env['PATH_INFO']
-        filename = "#{@config.resolve_swagger_root(env)}/#{path}"
+        # Sanitize the filename for directory traversal by expanding, and matching the swagger root directory
+        filename = File.expand_path(File.join(@config.resolve_swagger_root(env), path))
+        unless filename.match Regexp.new('^' + Regexp.escape(@config.resolve_swagger_root(env)))
```
stevegeek: Just a thought, could this just use 'start_with?', or if using regex should the expression maybe use '\A' rather than '^'?

sergioisidoro: `start_with?` is much more ruby like, I like it. While checking it also noticed that[ `expand_path` also does the file ](https://www.rubydoc.info/stdlib/core/File.expand_path) joining, so I can remove that File.join with the `dir_string` param.

I'll merge if tests pass after the changes.

stevegeek: ok thanks!

romanblanco (approved):

rswag-api/lib/rswag/api/middleware.rb
carlobeltrame: @sergioisidoro Unfortunately, the `dir_string` will be ignored in case the file name starts with a slash (as is the case with absolute paths, but also in our case of getting the path out of `env['PATH_INFO']`). So the "simplify file expansion" part broke the functionality of the gem. Will post a PR to fix this.

carlobeltrame: #660

PR #656, opened 2023-07-06
Improving generation of  OAS3.0 `requestBody` from `parameter`s

stevegeek:
## Problem / Solution

The handling of path `parameter`s does not currently generate OAS3.0 `requestBody` correctly in some cases. This PR proposes a refactoring and updates to `SwaggerFormatter` to improve translation of the DSL defined parameters. 

It takes the approach that any parameter defined as `in: :body` is handled differently to parameters that are `in: :formData`. A request can only have a single `body` parameter but can have multiple `formData` parameters.

It also adds support for translating a `:file`parameter to the new OAS3.0 schema and some tests for the http `bearer` security schema.

I started this as I wanted to specify parameter for multipart requestBody (including file upload with multiple possible content types) and ended up making these changes so far. 

I thought to now open a draft PR for any early feedback and then hopefully update until its good to go.

* [requestBody](https://swagger.io/docs/specification/describing-request-body/)
* [File upload](https://swagger.io/docs/specification/describing-request-body/file-upload/)

Note that file upload definitions improved in OAS3.1 but this PR does not yet attempt to support that

- [x] OAS2
- [x] OAS3
- [ ] OAS3.1

### Checklist
- [x] Added tests
- [ ] Changelog updated

---

Thanks so much for all the work on `rswag` over the years, its great!

### DIFF
--- a/README.md
+++ b/README.md
@@ -793,6 +793,9 @@ after do |example|
 end
 ```
 
+Note you will need to disable the `--dry-run` option for Rspec 3.0.0 or higher to 
+run the after callback. See below for more information.
+
 #### Dry Run Option ####
 
 The `--dry-run` option is enabled by default for Rspec 3, but if you need to
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -52,35 +52,14 @@ module Rswag
       def stop(_notification = nil)
         @config.swagger_docs.each do |url_path, doc|
           unless doc_version(doc).start_with?('2')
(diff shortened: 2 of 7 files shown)


romanblanco:
@stevegeek Thank you for the work!

I like to see the change in this area as it currently also troubles me, however, the rswag gem is (hopefully) soon going to drop the support for Swagger (aka OpenAPI v2), see https://github.com/rswag/rswag/pull/574.

Maybe it would make sense to base your changes on those in the PR to avoid re-doing the changes later.

stevegeek:
@romanblanco sure, will do!

stevegeek:
@romanblanco https://github.com/romanblanco/rswag/pull/1/files 😄

PR #660, opened 2023-07-13
Fix path expansion

carlobeltrame:
Reverts part of efd4ea4

## Problem
The dir_string is ignored if file_name starts with a slash. In our case of retrieving the HTTP request path from env, path will start with a slash. This broke swagger file resolution.

## Solution
Revert back to the `File.join` method which is safer in case of risk of additional slashes.

-

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
https://github.com/rswag/rswag/pull/654#discussion_r1262688116
#653

### Checklist
- [ ] Added tests
- [ ] Changelog updated


romanblanco (approved):

sergioisidoro:
Damn. We should add a regression test, because all tests passed :/

Sorry for that.

Issue #606, opened 2023-03-08
[BUG] '+' sign in url does not get encoded properly


NebojsaStrbac997:
## Describe the bug
I have a param in query called 'created_after'. It is a string with datetime format and when being sent, it isn't encoded properly. The plus sign turns into a 'space' in the URL. 

## Steps to Test or Reproduce
created_after is defined like this:

` parameter name: :created_after, in: :query, required: false, type: :string, format: :datetime `

My test looks like this: 
```
 response 200, "Get all examples that are created after 'created_after' value" do
  let(:old_example) { create(:example, created_at: Datetime.current - 5.days) }
  let(:created_after) { (old_example.created_at + 5.minutes).to_formatted_s.sub(" ", "T").gsub(" ", "") }
  
  run_test! do
    excluded_example_id = old_example.id
    expect(json_response_data_ids).to_not include(excluded_example_id)
   end
 end
```
When swaggerizing the tests, they fail. In my rails logs I notice that this happens:

```
Started GET "/api/v1/example/index?created_after=2023-03-05T15:33:01+0100"

Processing by Api::V1::example::exampleController#index as HTML Parameters: {"created_after"=>"2023-03-05T15:33:01 0100"
```
In Parameters: the plus sign is changed to a space, instead of being encoded back into a plus sign.

Here is the code for created_after in my controller:
```
      def created_after_param
        created_after = params.fetch(:created_after, nil)
        created_after.present? ? parse_created_after_param(created_after) : nil
      end

      def parse_created_after_param(created_after)
        Time.strptime(created_after, "%Y-%m-%dT%H:%M:%S%z")
      rescue ArgumentError
        nil
      end
```

## Expected behavior
In parameters, created_after should have a + at the end, and thus the test will pass.

## Dependency versions
* Rswag: 2.8.0
* RSpec: 3.11.0
* Rails: 7.0.3
* Ruby: 3.1.2

## Relates to which version of OAS (OpenAPI Specification)
- [ ] OAS2
- [ ] OAS3
- [x] OAS3.1

NebojsaStrbac997:
@romanblanco I have found a temporary workaround. Basically I've added CGI.escape to the string inside created after like this:

`let(:created_after) { CGI.escape((old_notice.created_at + 2.minutes).to_s.sub(" ", "T").gsub(" ", "")) }`

[...]

PR #671, opened 2023-08-07
Fix encoding of query params when openapi is v3

HiroNonoyama:
## Problem
URL encoding is not processed correctly when I define a string query parameter as `{ schema: { type: :string } }` in OpenAPI v3. This issue occurs even if I define the schema as simply `{ type: :string }` when I have multiple test cases. This is because `{ type: :string }` is automatically converted to `{ schema: { type: :string } }` after one test case finishes and before the next one starts.

## Solution
Update `build_query_string_part` and handle query parameter defined as `{ schema: { type: :string } }`

- [ ] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
https://github.com/rswag/rswag/issues/606#issuecomment-1664379381

### Checklist
- [x] Added tests
- [x] Changelog updated
I didn't update readme because this is bug fix

Add a query parameter with `{ schema: { type: :string } }` that is in datetime format. Before this PR, the '+' symbol in the parameters would be replaced with a whitespace after encoding.

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -173,6 +173,8 @@ module Rswag
                           end
               return "#{escaped_name}=" + value.to_a.flatten.map{|v| CGI.escape(v.to_s) }.join(separator)
             end
+          when :string
+            return "#{escaped_name}=#{CGI.escape(value.to_s)}"
           else
             return "#{name}=#{value}"
           end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- Fix encoding query parameter for OpenApi v3 (https://github.com/rswag/rswag/pull/671)
+
 ### Added
 
 ### Changed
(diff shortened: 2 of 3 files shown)


NebojsaStrbac997:
@romanblanco Checkout this PR, we had a discussion [here](https://github.com/rswag/rswag/issues/606#issuecomment-1665359813)

rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -173,6 +173,8 @@ def build_query_string_part(param, value, swagger_doc)
                           end
               return "#{escaped_name}=" + value.to_a.flatten.map{|v| CGI.escape(v.to_s) }.join(separator)
             end
+          when :string
+            return "#{escaped_name}=#{CGI.escape(value.to_s)}"
           else
             return "#{name}=#{value}"
```
BrianHawley, line 179: Why not just change this line instead? That way you'll never be adding unescaped values.

BrianHawley:
#725 fixes the same bug but does the change I suggested instead. It should be used instead.

HiroNonoyama:
@BrianHawley 
That's better.
The PR you referenced is on [milestone#3](https://github.com/rswag/rswag/milestone/3), which is planed to be released before the milestone of this PR. Therefore I'll close this PR

Issue #562, opened 2022-10-10
[BUG] RSwag 2.6.0 incorrectly encodes date-formatted string into YAML


bulgakke:
## Describe the bug
RSwag translates Ruby's `"2022-11-22"` string into YAML's `2022-11-22` (without quotes). Then (e. g. when opening the /api-docs page) the YAML parser seems to crash when trying to read it:
![image](https://user-images.githubusercontent.com/56848262/194936047-7420e88c-16e9-4a96-b646-689893a0bb59.png)

## Steps to Test or Reproduce
This code:
```ruby
parameter name: :params, in: :body, schema: {
  type: :object,
  properties: {
    date: {
      type: :string, 
      format: :date, 
      example: "2022-11-28"
    }
  }
}
```
generates this in `swagger.yaml`:
```yaml
# RSwag 2.6.0
requestBody:
  content:
    application/json:
      schema:
        type: object
        properties:
          date:
            type: string
            format: date
            example: 2022-11-28 # <=== The problem is here!
```

## Expected behavior
Using RSwag 2.5.1, the same code generates:
```yaml
# RSwag 2.5.1
requestBody:
  content:
    application/json:
      schema:
        type: object
        properties:
          date:
            type: string
            format: date
            example: '2022-11-28' # <=== Properly quoted
```
And Swagger docs page works as expected.

If I set `config.swagger_format = :json`, it also works as expected:
```json
"requestBody": {
  "content": {
    "application/json": {
      "schema": {
        "type": "object",
        "properties": {
          "date": {
            "type": "string",
            "format": "date",
            "example": "2022-11-28"
          }
        }
      }
    }
  }
}
```

Tokinh0:
Is this issue fixed in 2.8? Maybe I can deal with it, just need some direction.

ghost:
Not fixed as of 2.11.0
I think the issue related to https://github.com/ruby/psych/issues/262
Any example that includes a String that looks like a Date (i.e. formatted like 2022-11-28) will trigger this issue regardless of whether or not there is a `format: :date`

I think the solution as far as RSwag is concerned is to make Date a permitted class

asavageiv:
Unfortunately that PR does not work. Allowing the String to be loaded as a Date causes it to display incorrectly in the UI. It shows up as `{}` instead of the String "2022-11-28". 

I think we need to dump using `safe_dump` which requires `psych >= 4.0.1`

ghost:
[...]

PR #680, opened 2023-10-11
Permit loading the Date class when loading YAML files

asavageiv:
Workaround for rswag/rswag#562

## Problem
RSwag generates YAML with strings that look like Dates but are not. This breaks parsing because Dates are not a permitted class for loading.

## Solution
Permit loading the Date class

N/A

N/A

### Related Issues
#562
[ruby/psych/issues#262
](https://github.com/ruby/psych/issues/262)

### Checklist
- [X] Added tests
- [ ] Changelog updated

`bundle exec rspec rswag-api/spec/`

### DIFF
--- a/rswag-api/lib/rswag/api/middleware.rb
+++ b/rswag-api/lib/rswag/api/middleware.rb
@@ -48,7 +48,8 @@ module Rswag
       end
 
       def load_yaml(filename)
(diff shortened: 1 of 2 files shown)


asavageiv:
This does not work

PR #621, opened 2023-04-07
Fix encoding of query params

NebojsaStrbac997:
## Problem
In query parameters, date-time formatted strings would not be encoded properly. This PR aims to fix this issue, and could potentially fix more issues than that. Strings should be escaped and formatted properly when being sent as query parameters.

## Solution
Add CGI.escape to method build_query_string_part

### This concerns this parts of the OpenAPI Specification:
--

### The changes I made are compatible with:
- [ ] OAS2
- [x] OAS3
- [ ] OAS3.1

### Related Issues
https://github.com/rswag/rswag/issues/606

### Checklist
- [x] Added tests
- [x] Changelog updated
- [x] Added documentation to README.md

### Steps to Test or Reproduce
Add a parameter in query, that is a string type and is of datetime format.
Before this PR, the '+' symbol in the parameters would be turned into a white space after encoding.

PR #683, opened 2023-10-12
Factor out query param serialization into classes

ClaytonPassmore:
## Problem

The query parameter serialization logic in the `RequestFactory` class is complicated and difficult to test well. In addition, it does not consistently escape query parameter names and values (depends on whether you're using OAS 2 vs OAS 3).

## Solution

Split out the parameter serialization logic into classes where logic can be reused and more easily unit tested.

### Notes

This change was made with the intention of being 100% compatible with the previous implementation. The only exception is that OAS 2 query parameters are now escaped with `CGI.escape` when previously they were not.

If the change is an issue, I can tweak my changes to preserve the previous escaping behaviour.

### File Structure

I added a new `query_serializers` folder with the following structure:

```
rswag-specs/lib/rswag/specs/query_serializers
├── collections
│   ├── multi_serializer.rb
│   └── xsv_serializer.rb
├── primitive_serializer.rb
└── specifications
    ├── oas3_serializer.rb
    └── swagger_serializer.rb
```

* The `collections` folder contains serializers that work on arrays.
* The `specifications` folder contains serializers that adhere to specifications like OAS 2 or OAS 3

I'm definitely open to changing class names and/or moving files around!

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
Links to any related issues.

#621 fixed query parameter encoding for OAS 3, but that was never back-ported to OAS 2 serialization.

### Checklist
- [x] Added tests
- [ ] Changelog updated
- [ ] ~Added documentation to README.md~ N/A
- [ ] ~Added example of using the enhancement into [test-app](https://github.com/rswag/rswag/tree/master/test-app)~ N/A

### Steps to Test

The existing specs were left to show that I didn't break compatibility with the original implementation. I added unit tests for each added class to ensure 100% coverage.

### Motivation

While I truly believe this code is more maintainable, I didn't make this change without selfish motivations. 😄 

I currently have a fork of this gem where I have made changes to support a custom query parameter format. By splitting the logic out into classes, I can change my application to monkey patch specific classes to support my custom format without needing to maintain a fork.

### DIFF
--- a/rswag-specs/lib/rswag/specs/query_serializers/collections/multi_serializer.rb
+++ b/rswag-specs/lib/rswag/specs/query_serializers/collections/multi_serializer.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Rswag
+  module Specs
+    module QuerySerializers
+      module Collections
+        class MultiSerializer
+          def serialize(name, value)
+            value.map { |v| "#{CGI.escape(name.to_s)}=#{CGI.escape(v.to_s)}" }.join("&")
+          end
+        end
+      end
+    end
+  end
+end
--- a/rswag-specs/lib/rswag/specs/query_serializers/collections/xsv_serializer.rb
+++ b/rswag-specs/lib/rswag/specs/query_serializers/collections/xsv_serializer.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Rswag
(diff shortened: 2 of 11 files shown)


ClaytonPassmore:
Didn't realize spelling mattered. Hopefully that latest change gets CI passing.

PR #686, opened 2023-10-20
fix: remove invalid fields when generating oas 3 docs

Eddayy:
## Problem
Rswag::Specs::SwaggerFormatter supports default content type when you defined consumes or produces in swagger_doc but this only works for OAS 2 as it adds field that are incompatible with OAS 3

## Solution
Remove invalid root fields when generating for OAS 3 

I remove fields that do not belong in https://spec.openapis.org/oas/v3.1.0#fixed-fields

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
Links to any related issues.

### Checklist
- [x] Added tests
- [ ] Changelog updated

Add `produces:` or any key in the swagger_docs config, will remove any invalid fields if openapi is version 3

### DIFF
--- a/rswag-specs/lib/rswag/specs/swagger_formatter.rb
+++ b/rswag-specs/lib/rswag/specs/swagger_formatter.rb
@@ -80,10 +80,13 @@ module Rswag
 
                   value[:parameters].reject! { |p| p[:in] == :body || p[:in] == :formData }
                 end
-                remove_invalid_operation_keys!(value)
+                remove_invalid_path_operation_keys!(value)
               end
             end
+
+            remove_invalid_root_keys!(doc)
           end
+
 
           file_path = File.join(@config.swagger_root, url_path)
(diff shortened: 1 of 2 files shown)


joshraker:
~For those looking for a workaround in the meantime, I've had success with adding an `after(:spec)` block to the `swagger_helper` config~

```ruby
RSpec.configure do |config|
  ...
  config.after(:suite) do
    spec.delete(:consumes)
    spec.delete(:produces)
  end
end
```

Edit: The suggestion above broke the OpenAPI spec generation. No `requestBody` blocks were generated.

jtannas:
Hi @Eddayy  - apologies but I haven't been paying attention to this repo for a long time.
We're currently working on `v3` branch that drops support for OAS 2 (see #714) and the bulk of our development is going on there.

Could you take a look at that branch and see if your work applies there?
Thanks :bow:

Issue #674, opened 2023-09-11
[BUG] Example group metadata does not support symbol shorthand


franzliedke:
## Describe the bug
The custom example groups in RSwag (for example `response`) do not support [symbol-style example metadata](https://rspec.info/features/3-12/rspec-core/metadata/user-defined/).

## Steps to Test or Reproduce
Example spec:

```ruby
response "201", "article created", :skip_bullet do
  # ...
end
```

This is perfectly valid in RSpec, but leads to the following error inside RSwag:

```
NoMethodError:
  undefined method `[]=' for :skip_bullet:Symbol
```

## Expected behavior

The spec should be equivalent to the following, which works just fine:

```ruby
response "201", "article created", skip_bullet: true do
  # ...
end
```

## Additional context
This became a problem with the new `MetadataStyle` cop in rubocop-rspec, which has support for two enforced styles, `hash` or `symbol`.

Rubocop understands that `response` defines an example group with the following config:
```yaml
RSpec:
  Language:
    ExampleGroups:
      Regular:
        - response
```

## Dependency versions
The version of are you using for:
* Rswag: 2.9.0
* RSpec: 3.12
* Rails: 7.0.8
* Ruby: 3.2.2

## Relates to which version of OAS (OpenAPI Specification)
Not related to OAS

perfetti:
This seems a user issue, symbol-style example metadata which is `:foo => :bar`, as that is shorthand for a hash.  `{ foo: }` could also be acceptable as a hash assuming you've defined a value for `foo` earlier in the code.

You'll notice in the link you posted the syntax is `:foo => :bar`

Using `:foo` means you're passing purely a symbol and not a hash. Hence the error.

PR #687, opened 2023-10-25
Allowing response metadata to also be passed in in the symbol metadata style.

perfetti:
## Problem

https://github.com/rswag/rswag/issues/674

In keeping with metadata styling, as per rubocop's [rspec metadata style](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecmetadatastyle). We should be able to pass

## Solution
Handle the case of metadata being passed as a symbol.

- [X] OAS2
- [X] OAS3
- [X] OAS3.1

### Related Issues
Links to any related issues.

### Checklist
- [x] Added tests
- [ ] Changelog updated

### DIFF
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -67,6 +67,9 @@ module Rswag
       end
 
       def response(code, description, metadata = {}, &block)
+        if metadata.is_a?(Symbol)
+          metadata = { metadata => true }
+        end
         metadata[:response] = { code: code, description: description }
         context(description, metadata, &block)
       end
--- a/rswag-specs/lib/rswag/specs/example_group_helpers.rb
+++ b/rswag-specs/lib/rswag/specs/example_group_helpers.rb
@@ -7,14 +7,18 @@ module Rswag
     module ExampleGroupHelpers
       ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: Support for Ruby 2.6 will be dropped in v3.0') if RUBY_VERSION.start_with? '2.6'
 
-      def path(template, metadata = {}, &block)
+      def path(template, *metadata, &block)
+        metadata = _parse_metadata(metadata)
         metadata[:path_item] = { template: template }
         describe(template, metadata, &block)
       end
 
       [:get, :post, :patch, :put, :delete, :head, :options, :trace].each do |verb|
-        define_method(verb) do |summary, **metadata, &block|
-          api_metadata = { operation: { verb: verb, summary: summary } }.deep_merge(metadata)
+        define_method(verb) do |summary, *metadata, &block|
+          metadata = _parse_metadata(metadata)
+          api_metadata =
+            { operation: { verb: verb, summary: summary } }
+            .deep_merge(metadata)
           describe(verb, **api_metadata, &block)
         end
       end
@@ -67,14 +71,7 @@ module Rswag
       end
 
       def response(code, description, *metadata, &block)
-        metadata = metadata.reduce({}) do |acc, item|
-          if item.is_a?(Symbol)
-            acc[item] = true
-          elsif item.is_a?(Hash)
-            acc.merge!(item)
-          end
-          acc
-        end
+        metadata = _parse_metadata(metadata)
         metadata[:response] = { code: code, description: description }
         context(description, metadata, &block)
       end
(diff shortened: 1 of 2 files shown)


rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
franzliedke, line 1: Should this handle multiple symbols as well?

```ruby
subject.response('201', 'success', :foo, :bar)
```

And now that I think about it, a mix of Boolean and Symbol parameters would be a realistic case, too:

```ruby
subject.response('201', 'success', :sidekiq, freeze_time: '2021-01-01')
```

franzliedke (changes requested):
Thanks for tackling this! 💚

rswag-specs/spec/rswag/specs/example_group_helpers_spec.rb
perfetti: Yeah, I think that makes sense. I'll tackle that tonight.

perfetti: Or I'll tackle it now while waiting on the bus.

franzliedke (changes requested):
This implementation looks more complete, thanks!

However, I just realized that this change should ideally be applied to all DSL methods that delegate to `describe` or `context`. Should have made that clearer in the ticket, sorry. 😬

perfetti:
@franzliedke Added that functionality. Let me know if I missed anywhere.

rswag-specs/lib/rswag/specs/example_group_helpers.rb
```diff
@@ -149,6 +154,19 @@ def run_test!(description = nil, *args, **options, &block)
...
+      def _parse_metadata(metadata)
+        metadata.reduce({}) do |acc, item|
+          if item.is_a?(Symbol)
+            acc[item] = true
+          elsif item.is_a?(Hash)
+            acc.merge!(item)
+          end
+          acc
+        end
+      end
```
franzliedke, line 169: Mini nitpick - this avoids the need for returning something from the block: 🤷🏼 

```suggestion
      def _parse_metadata(metadata)
        metadata.each_with_object({}) do |item, acc|
          if item.is_a?(Symbol)
            acc[item] = true
          elsif item.is_a?(Hash)
            acc.merge!(item)
          end
        end
      end
```

rswag-specs/lib/rswag/specs/example_group_helpers.rb
```diff
@@ -7,14 +7,18 @@ module Specs
...
+      def path(template, *metadata, &block)
+        metadata = _parse_metadata(metadata)
         metadata[:path_item] = { template: template }
         describe(template, metadata, &block)
       end
 
       [:get, :post, :patch, :put, :delete, :head, :options, :trace].each do |verb|
-        define_method(verb) do |summary, **metadata, &block|
-          api_metadata = { operation: { verb: verb, summary: summary } }.deep_merge(metadata)
+        define_method(verb) do |summary, *metadata, &block|
```
franzliedke, line 17: I wonder... this kinda rebuilds RSpec's handling of parameters. Judging by [`describe`'s signature](https://github.com/rspec/rspec-core/blob/1eeadce5aa7137ead054783c31ff35cbfe9d07cc/lib/rspec/core/example_group.rb#L247-L275) in rspec-core, this could also work if we just pass on `*metadata` as second parameter to `describe`. Do the tests still pass if you do that?

franzliedke (approved):
I think you handled all the cases, and the tests seem to describe the behavior I would like to see. So from my side this is an approval (but I have no power anyway ^^).

I also added a suggestion, though: I think we can leave the transformation details to RSpec (see inline comment).

perfetti:
@BookOfGreg Any thoughts on this PR?

rswag-specs/lib/rswag/specs/example_group_helpers.rb
BrianHawley: @franzliedke I don't know about tests failing, but that approach would fail if the extra metadata provided by the user was in hash form.

However, you don't need to parse it. Since none of rswag's metadata is in the set-if-true style, all you would need to do is something like this:
```ruby
define_method(verb) do |summary, *args, &block|
  metadata = args.last.is_a?(Hash) ? args.pop : {}
  api_metadata = { operation: { verb: verb, summary: summary } }.deep_merge(metadata)
  describe(verb, *args, **api_metadata, &block)
end
```
No parsing needed.

pboling:
RSwag team is missing.  Maybe it needs more members who can at least keep simple triage tasks going.

pboling:
@perfetti Did you find an alternative project to use instead of rswag?

perfetti:
@pboling I kept using rswag.

pboling:
@perfetti why close the PR?

perfetti:
@pboling I didn't close the PR, I think I deleted the fork by accident with a stale request

perfetti:
@pboling I've reopened this over at https://github.com/rswag/rswag/pull/796

PR #702, opened 2023-11-26
Fix deprecation warnings in version 2.12.0

### LINKED ISSUE #703: [BUG] Depreciation warning adding noise to test runs
## Describe the bug

Usefulness of new deprecation warnings added. 

## Steps to Test or Reproduce

In https://github.com/rswag/rswag/pull/688, deprecation warnings were added to various methods which the gem itself depends on and it causes the warnings to be logged in test runs adding noise to the output. 

## Expected behavior

Since there is no replacement for the existing methods which will be renamed, I don't think there is a need to print out the deprecation message since there is nothing the end user can do. 

## Dependency versions
The version of are you using for:
* Rswag: 2.12.0
* RSpec: 3.12.0
* Rails: 7
* Ruby: 3.2

hasghari:
## Problem
Version 2.12.0 added deprecation warnings to several configuration options without adding the alternative options to resolve the warnings.

## Solution
This pull request uses `ActiveSupport::Deprecation#deprecate_methods` to correctly warn users about the deprecated settings on the `RSpec` configuration object.

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Checklist
- [x] Added tests
- [x] Changelog updated

### DIFF
--- a/rswag-specs/lib/rswag/specs/configuration.rb
+++ b/rswag-specs/lib/rswag/specs/configuration.rb
@@ -7,63 +7,59 @@ module Rswag
         @rspec_config = rspec_config
       end
 
-      def swagger_root
-        ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: The method will be renamed to "openapi_root" in v3.0')
-        @swagger_root ||= begin
-          if @rspec_config.swagger_root.nil?
-            raise ConfigurationError, 'No swagger_root provided. See swagger_helper.rb'
+      def openapi_root
+        @openapi_root ||=
+          @rspec_config.openapi_root || raise(ConfigurationError, 'No openapi_root provided. See swagger_helper.rb')
+      end
+
+      def openapi_specs
+        @openapi_specs ||= begin
+          if @rspec_config.openapi_specs.nil? || @rspec_config.openapi_specs.empty?
+            raise ConfigurationError, 'No openapi_specs defined. See swagger_helper.rb'
           end
 
-          @rspec_config.swagger_root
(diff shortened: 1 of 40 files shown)


alexcastrodev:
Good, i was checking here about the updates :D

romanblanco:
@hasghari Thanks a lot for putting the PR together! 

I'll try to test and merge this soon.

rswag-specs/lib/rswag/specs/configuration.rb
```diff
@@ -7,63 +7,59 @@ def initialize(rspec_config)
         @rspec_config = rspec_config
       end
 
-      def swagger_root
-        ActiveSupport::Deprecation.warn('Rswag::Specs: WARNING: The method will be renamed to "openapi_root" in v3.0')
-        @swagger_root ||= begin
-          if @rspec_config.swagger_root.nil?
-            raise ConfigurationError, 'No swagger_root provided. See swagger_helper.rb'
+      def openapi_root
```
romanblanco, line 10: @hasghari, the `swagger_root` in https://github.com/rswag/rswag/blob/8514da308f11d393fdf15c1e90be11061530a5d2/rswag-api/lib/rswag/api/configuration.rb#L8 should be updated too

hasghari: @romanblanco Thanks for the feedback. I've added deprecation warnings for the `swagger_root` setting in the `rswag-api` configuration.

romanblanco (approved):
Tested, works :+1:

PR #722, opened 2024-01-08
Apply CGI.escape to non-array non-form values

krainboltgreene:
## Problem
A string query param described in an rswag integration spec gets escaped in `rspec` tests and doesn't get escaped in rswag non-dry run.

## Solution
Apply the same array and form escaping to everything else.

- [X] OAS2
- [X] OAS3
- [X] OAS3.1

### Checklist
- [ ] Added tests
- [ ] Changelog updated

1. Have a query param called `x` that is a string and contains a `#`, for example: `?filter[description]=Order for #231`
2. Run the test and inspect the params in the controller

### DIFF
--- a/rswag-specs/lib/rswag/specs/request_factory.rb
+++ b/rswag-specs/lib/rswag/specs/request_factory.rb
@@ -174,7 +174,7 @@ module Rswag
               return "#{escaped_name}=" + value.to_a.flatten.map{|v| CGI.escape(v.to_s) }.join(separator)
             end
           else
-            return "#{name}=#{value}"
+            return "#{name}=#{CGI.escape(value)}"
           end
         end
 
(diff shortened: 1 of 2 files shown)


rswag-specs/lib/rswag/specs/request_factory.rb
```diff
@@ -174,7 +174,7 @@ def build_query_string_part(param, value, swagger_doc)
               return "#{escaped_name}=" + value.to_a.flatten.map{|v| CGI.escape(v.to_s) }.join(separator)
             end
           else
-            return "#{name}=#{value}"
+            return "#{name}=#{CGI.escape(value)}"
```
BrianHawley, line 177: `CGI.escape(value.to_s)`
The `to_s` was implicit before.

BrianHawley:
#725 fixes the same bug but does the change I suggested instead. It should be used instead.

PR #723, opened 2024-01-10
Improving generation of  OAS3.0 `requestBody` from `parameter`s

stevegeek:
## Problem / Solution

The handling of path `parameter`s does not currently generate OAS3.0 `requestBody` correctly in some cases. This PR proposes a refactoring and updates to `SwaggerFormatter` to improve translation of the DSL defined parameters. 

It takes the approach that any parameter defined as `in: :body` is handled differently to parameters that are `in: :formData`. A request can only have a single `body` parameter but can have multiple `formData` parameters.

It also adds support for translating a `:file`parameter to the new OAS3.0 schema and some tests for the http `bearer` security schema.

I started this as I wanted to specify parameter for multipart requestBody (including file upload with multiple possible content types) and ended up making these changes so far. 

* [requestBody](https://swagger.io/docs/specification/describing-request-body/)
* [File upload](https://swagger.io/docs/specification/describing-request-body/file-upload/)

Note that file upload definitions improved in OAS3.1 but this PR does not yet attempt to support that

- [x] OAS3
- [ ] OAS3.1

### Checklist
- [x] Added tests
- [ ] Changelog updated

---

Thanks so much for all the work on `rswag` over the years, its great!


rswag-specs/lib/rswag/specs/openapi_formatter.rb
```diff
@@ -158,6 +129,120 @@ def remove_invalid_operation_keys!(value)
...
+          end
+        end
+      end
+
+      def parse_endpoint(endpoint, mime_list)
+        parameters = endpoint[:parameters]
+        # There can only be 1 body parameter in Swagger 2.0, so while in OAS3 we interpret
+        # body parameters as formData, only consider the first body we encounter.
+        schema_param = parameters.find { |p| p[:schema] if p[:in] == :body }
+        parse_body_parameter(endpoint, schema_param, mime_list) if schema_param
```
romanblanco: @stevegeek, the Swagger 2.0 support is being dropped in this branch.
It seems to me that the `parse_body_parameter` function can be dropped.

stevegeek:
@romanblanco hey Roman, anything else needed here to get this merged into `v3`?

stevegeek:
@romanblanco updated the branch to reflect `request_*` changes. I merged this into `v3` with #774 and tests pass it seems

(if/when #774 is merged I will update this branch again to resolve conflicts)

romanblanco:
@stevegeek Updating v3 branch in https://github.com/rswag/rswag/pull/784 since commit has been merged into master again since.

stevegeek:
@romanblanco ok this has been updated

rswag-specs/lib/rswag/specs/openapi_formatter.rb
```diff
@@ -164,6 +130,116 @@ def remove_invalid_operation_keys!(value)
...
+            remove_invalid_operation_keys!(endpoint)
+          end
+        end
+      end
+
+      def parse_endpoint(endpoint, mime_list)
+        parameters = endpoint[:parameters]
+        # But there can be any number of formData
+        parameters.select { |p| p[:schema] if p[:in] == :formData  }.each do |schema_param|
+          parse_parameter_schema(endpoint, schema_param, mime_list)
```
romanblanco: @stevegeek, I'm testing the changes in https://github.com/RedHatInsights/compliance-backend/compare/master...romanblanco:compliance-backend:rswag-v3 and have a problem that response examples are not being generated.

The `parse_parameter_schema` method is never called, so nor is the `set_mime_examples`.

romanblanco: Nevermind, had a bad setting for `RSWAG_DRY_RUN`.

romanblanco:
@stevegeek One issue I'm running into when testing is that `body` params are generated in between other parameters, rather than inside of `requestBody`, rendering invalid spec.

(see 
https://github.com/romanblanco/compliance-backend/blob/85b8be8fc35230ce43cb9f8f7c71488a954814ce/spec/integration/v2/policies_spec.rb#L85

https://github.com/romanblanco/compliance-backend/blob/85b8be8fc35230ce43cb9f8f7c71488a954814ce/openapi/v2/openapi.json#L475)

I'm trying to identify cause of this, to confirm whether the problem is on my side, introduced in this PR or already in `v3` branch.

stevegeek:
@romanblanco Ok sure, ill see if I can see what's wrong. Prob need to review this myself as I wrote it along time ago. Btw from the examples above, should it be `in: :formData` instead of `in: :body` ?

romanblanco:
Trying `:formData`, the [parameter data](https://github.com/romanblanco/compliance-backend/blob/85b8be8fc35230ce43cb9f8f7c71488a954814ce/spec/integration/v2/policies_spec.rb#L89) are not getting into the request and it fails.

`:body` has been used originally ([diff](https://github.com/RedHatInsights/compliance-backend/compare/master...romanblanco:compliance-backend:rswag-v3#diff-37a69882b055bd410152b110b31a2a43e83270cb415edd07d5baa3801aef4965R85)) and should be supported.

test-app/openapi/v1/openapi.json
```diff
@@ -89,7 +113,14 @@
         "description": "Creates a new blog from provided data",
         "operationId": "createBlog",
         "parameters": [
-
+          {
+            "name": "blog",
```
romanblanco, line 117: @stevegeek The [issue](https://github.com/rswag/rswag/pull/723#issuecomment-2410490552) we're discussing in thread is also happening here.

test-app/openapi/v1/openapi.json
```diff
@@ -89,7 +113,14 @@
...
+            "in": "body",
```
romanblanco: @stevegeek The [issue](https://github.com/rswag/rswag/pull/723#issuecomment-2410490552) we're discussing in thread is also happening here.

stevegeek:
Ok sure. `:body` can be supported as per the PR. But I would consider deprecating it as its not really a thing in OAS 3+ as far as I understand so could be confusing: https://swagger.io/docs/specification/v3_0/describing-request-body/describing-request-body/?sbsearch=body  "Body and form parameters are replaced with requestBody."

stevegeek:
> Trying `:formData`, the [parameter data](https://github.com/romanblanco/compliance-backend/blob/85b8be8fc35230ce43cb9f8f7c71488a954814ce/spec/integration/v2/policies_spec.rb#L89) are not getting into the request and it fails.
> 
> `:body` has been used originally ([diff](https://github.com/RedHatInsights/compliance-backend/compare/master...romanblanco:compliance-backend:rswag-v3#diff-37a69882b055bd410152b110b31a2a43e83270cb415edd07d5baa3801aef4965R85)) and should be supported.

OK interesting. Im using this branch with `v3` and its works for me with this for examples:

```
    patch("update a user") do
      tags "account/users"

      consumes "multipart/form-data"
      produces "application/json"

      parameter name: "email", in: :formData, type: :string, description: "User's email address (this is required by the user)."
      parameter name: "first_name", in: :formData, type: :string, description: "User's first name."
      parameter name: "last_name", in: :formData, type: :string, description: "User's last name."
      parameter name: "position_name", in: :formData, type: :string, description: "User's position in the organisation."

      security [apiBearerToken: []]

      let(:request_params) { {"email" => email, "first_name" => first_name, "last_name" => last_name, "position_name" => position_name, "id" => user.encoded_id} }
```

romanblanco:
> I would consider deprecating it as its not really a thing in OAS 3+

I agree.

test-app/openapi/v1/openapi.json
stevegeek: Yeah ok i see, working on it now

romanblanco:
> OK interesting. Im using this branch with v3 and its works for me

I think you might be right. I'm also able to fix some of them with using `:formData`.
Probably the problem is on my side than.

stevegeek:
Pushed a small change to allow parsing `in: :body` in same way as `:formData`

romanblanco:
Thanks for the update, @stevegeek 

I'm now noticing only this behavior change:

```diff
        "requestBody": {
          "content": {
            "application/vnd.api+json": {
              "schema": {
-               "$ref": "#/components/schemas/policy"
+               "type": "object",
+               "properties": {
+                 "data": {
+                   "$ref": "#/components/schemas/policy"
+                 }
+               }
              }
            }
          }
```
before: 
image
after:
image

test-app/openapi/v1/openapi.json
```diff
@@ -204,14 +234,19 @@
...
               "schema": {
-                "oneOf": [
-                  {
-                    "$ref": "#/components/schemas/blog"
-                  },
-                  {
-                    "$ref": "#/components/schemas/flexible_blog"
+                "type": "object",
+                "properties": {
+                  "flexible_blog": {
```
romanblanco: @stevegeek I believe the `type` nor `properties` should not be there.

It renders the request example including the `flexible_blog`, which is changes the content from what is expected.

stevegeek: ok will take a look. Probably Im mixing up between form data fields and the body as a whole which is what is expected here

stevegeek:
@romanblanco so these latest changes should fix the issues and the output OpenAPI spec from the test-app integration tests seems to validate now correctly . I removed a spec suite which was from early OAS V3 support but it was only partially implemented and relied on stubs/mocks that made it quite annoying to work with. I think better to use actual `test-app` specs and the `OpenapiFormatter` specs to test things anyway. Can you take a look vs your code base?

romanblanco:
Thanks for the update @stevegeek! I'll get through it as soon as I can.

romanblanco:
@stevegeek I'm noticing that after latest change, rswag generates extra `description` attribute into `schema`.

e.g.

```diff
             "required": false,
             "description": "Indicates whether to return only resource IDs.",
             "schema": {
-              "type": "boolean"
+              "type": "boolean",
+              "description": "Indicates whether to return only resource IDs."
             }
           },
           {
```

rswag-specs/lib/rswag/specs/openapi_formatter.rb
```diff
@@ -190,9 +181,16 @@ def parse_formdata_or_body_parameter(endpoint, parameter, mime_list)
           set_mime_config(mime_config, parameter)
           set_mime_examples(mime_config, endpoint)
           parse_parameter_required(endpoint, mime_config, parameter)
+
+          parameter[:schema][:description] = parameter[:description] if parameter[:description]
```
romanblanco: This is the cause of difference from the current behavior.

stevegeek: Thanks, yeah I can't remember why that line was in the code before. I moved it hence the change of behaviour probably, but I'll have a look at why it's even there in the first place. This code is all very hard to maintain btw, cause it relies on mutating the options that come from the spec DSL into the OpenAPI format and that mutating is hard to reason about, would be nice to eventually rewrite this so the openAPI output is generated from the ground up from the input options

romanblanco: I agree it is a bit of maze to go through it.

I'm very happy for the refactoring done in this PR and take it as the first step in cleaning up the parsing in future.

stevegeek: Ok hopefully now fixed!

romanblanco (approved):
Looks good! :+1: 

Thanks @stevegeek

PR #741, opened 2024-04-09
Add support for OAS 3.0 oneOf discriminators

simba909:
## Problem
When adding an OpenAPI Discriminator Object to my response schema, rswag doesn't use it to disambiguate between similar schemas specified inside of a `oneOf`. This makes it impossible to express multiple schemas that are potentially ambiguous to the validator.

## Solution
Implement support for using a Discriminator Object (if available) to disambiguate between schemas to use for response validation.

Important: I'm not a Ruby developer by trade which means that this code likely can be improved by someone who knows what they're doing 😅  Any and all suggestions are welcome! I'd love to learn more about this. I've opted not to update the changelog or add documentation yet since I'd like to get feedback on the implementation first.

* [Discriminator Object](https://spec.openapis.org/oas/v3.1.0#discriminator-object)

- [ ] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
Links to any related issues.

### Checklist
- [x] Added tests
- [ ] Changelog updated

I've added test cases to validate that the library now uses the discriminator object.

### DIFF
--- a/rswag-specs/lib/rswag/specs/extended_schema.rb
+++ b/rswag-specs/lib/rswag/specs/extended_schema.rb
@@ -1,12 +1,14 @@
 # frozen_string_literal: true
 
 require 'json-schema'
+require 'rswag/specs/openapi_oneof'
 
 module Rswag
   module Specs
     class ExtendedSchema < JSON::Schema::Draft4
       def initialize
         super
+        @attributes['oneOf'] = OpenAPIOneOfAttribute
         @uri = URI.parse('http://tempuri.org/rswag/specs/extended_schema')
         @names = ['http://tempuri.org/rswag/specs/extended_schema']
       end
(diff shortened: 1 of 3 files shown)


simba909:
@romanblanco Unsure if you're the right person to request review from but maybe you can point me in the right direction 🙂

simba909:
@romanblanco Judging by recent work happening on the repo, should I change the base to the v3 branch instead of master? Given that it's OAS 3+

romanblanco:
@simba909 I'll probably get back to this PR once `v3` is merged in `master` and released.

I think you can keep the base, unless you prefer to already do the rebase now and test the changes with `v3`. I can always change it back to `master` once `v3` is out.

collink-root:
@romanblanco is this still something that's on your radar?

jonnyboyC:
Howdy just wanted to voice that we're also looking for this to land. We've got a deeply nested data structure and without something like this the test simply timesout. From a reading it sounds like we have a V3 branch that needs to land before this lands. 

@romanblanco If there is any way I can help contribute to make that happen I'd be happy to lend support

BookOfGreg:
We're always happy for contributors! Especially ones capable of writing clear, readable tests. :pray:

PR #749, opened 2024-05-09
Update generated rswag_ui.rb to use openapi_endpoint.

simi:
## Problem
Recently I have created new app and installed RSwag and it emits deprecation warning by default.

## Solution
Change template for generated `rswag_ui.rb` initializer to use `openapi_endpoint` instead of deprecated `swagger_endpoint`.

### Checklist
- [ ] Added tests (there is no test related per my understanding)
- [ ] Changelog updated (should I add entry?)

Run `rails g rswag:install` in new app.


simi:
fixed by https://github.com/rswag/rswag/commit/8fdb4f3125447b477207710612f56b707b5e40f4#diff-7b3140d65eef0fdd903988ebd7767734b26f58c5ccd4b59eba97bbf66d9b5a60R7

PR #639, opened 2023-05-19
Add support for text/plain body format

Ksm125:
## Problem
Unable to send request "string" body format correctly because it get serialized as JSON even if the content type is `text/plain`

## Solution
- Serialize to `JSON` only parameter body payload if  consumes  `application/json`

### This concerns this parts of the OpenAPI Specification:
* [EXAMPLE_LINK TO RELEVANT OPEN API SPECS PAGE](https://spec.openapis.org/oas/v3.1.0#data-types)
* [ANOTHER LINK TO RELEVANT OPEN API SPECS PAGE](https://spec.openapis.org/oas/v3.1.0#schema)

### The changes I made are compatible with:
- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
- https://github.com/rswag/rswag/issues/181

### Checklist
- [x] Added tests
- [x] Changelog updated
- [ ] Added documentation to README.md
- [ ] Added example of using the enhancement into [test-app](https://github.com/rswag/rswag/tree/master/test-app)

### Steps to Test or Reproduce
Outline the steps to test or reproduce the PR here.

PR #769, opened 2024-08-13
fix: Allow vendor-specific MIME types for JSON

### LINKED ISSUE #768: Improve check for JSON content type
Current check is too simplistic to consider all valid json types. For example, the JSON:API content type is [application/vnd.api+json](http://www.iana.org/assignments/media-types/application/vnd.api+json)

_Originally posted by @hasghari in https://github.com/rswag/rswag/pull/639#discussion_r1715933244_

hasghari:
## Problem
#639 added a bug that does not encode vendor-specific MIME types such as `application/vnd.api+json` as JSON payloads.

## Solution
Use a regular expression instead of the default `application/json` to detect JSON payloads.

* [Media Types](https://spec.openapis.org/oas/v3.1.0#media-types)
- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
Links to any related issues.

### Checklist
- [x] Added tests
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- Allow vendor-specific MIME types for JSON payloads (https://github.com/rswag/rswag/pull/769)
+
 ## [2.14.0] - 2024-08-13
 
 ### Added
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ case rails_version.split('.').first
 when '5'
   gem 'sqlite3', '~> 1.3.6'
 when  '6', '7'
-  gem 'sqlite3', '~> 1.4.1'
+  gem 'sqlite3', '~> 1.4'
 end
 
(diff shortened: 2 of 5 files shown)


hasghari:
I fixed the failing spec.

kwent:
Can we merge and ship a release? 🙏🏻

hasghari:
@romanblanco Removing the `mini_racer` gem seems to allow all the specs to pass. It's likely a stale dependency.

hasghari:
@romanblanco Can you re-run the failing github actions by busting the cache first? I verified that all specs were passing in my fork as you can see here: https://github.com/hasghari/rswag/actions/runs/10412163038

The only difference I see is that there's a cache hit in the failing specs whereas in my fork there was no cache hit.

thykel:
👋 Kind bump on this one, our JSON:API apps would be grateful!

gsar:
@romanblanco @hasghari i'm running into the same issue with `Content-Type: application/scim+json`. can we make the regex in the fix more generic please? something like `content_type =~ /\Aapplication\/(.+\+)?json\z/` is more appropriate.

PR #771, opened 2024-08-15
ci: Use appraisal gem for testing locally

hasghari:
## Problem
It's cumbersome to run the tests locally for all supported versions.

## Solution
Using the `appraisal` gem, we can run tests against all supported versions easier.

### Checklist
- [ ] Added tests
- [ ] Changelog updated


Gemfile
```diff
@@ -2,52 +2,28 @@
...
-  gem 'test-unit'
-  gem 'simplecov', '=0.21.2'
-end
-
-group :development do
-  gem 'rubocop'
-end
-
-group :assets do
-  gem 'mini_racer'
```
hasghari: I've removed the `mini_racer` gem as it was causing intermittent failures and it seems like the project no longer requires it as a dependency.

Issue #744, opened 2024-04-17
[BUG] Duplicate CSP headers when Rails CSP is configured


jboler:
## Describe the bug
rswag-ui returns duplicate CSP headers when Rails has a CSP configured.

## Steps to Test or Reproduce
Example repo: https://github.com/jboler/rswag-csp

The repo is an API only Rails 7.1 app with latest rswag.

To reproduce

```shell
bundle install
bin/rails s
curl -v localhost:3000/api-docs/index.html
```
Returns:
```
* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:3000...
* Connected to localhost (::1) port 3000
> GET /api-docs/index.html HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/html
< Content-Security-Policy: default-src 'self'; img-src 'self' data: https://validator.swagger.io; font-src 'self' https://fonts.gstatic.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; script-src 'self' 'unsafe-inline'; 
< content-security-policy: default-src 'self'
< etag: W/"113226cfb0f49141d46dc22487e30e76"
< cache-control: max-age=0, private, must-revalidate
< x-request-id: 3dfc9e63-3a84-4983-97f3-272c33b8c70a
< x-runtime: 0.003001
< server-timing: process_middleware.action_dispatch;dur=0.75
< Content-Length: 4173
< 
<!-- HTML for static distribution bundle build -->

```
Note the duplicate CSP headers. 

## Expected behavior
Ideally the rswag CSP header should override the rails CSP header.

I noticed that if I change the rswag-ui middleware source to use an all lower case `content-security-policy` header like rails then there is no duplicate header.

## Dependency versions
The version of are you using for:
* Rswag: 2.13.0
* Rails: 7.1.3.2
* Ruby: 3.3

eric1234:
[...]

PR #773, opened 2024-08-23
Fix CSP header name based on Rack version

SebouChu:
## Problem
Rack 3 now sends headers with lowercase names. This breaks the override CSP used by rswag, as the middleware writes the CSP with the name `Content-Security-Policy`, but Rails (with Rack 3) searches for `content-security-policy` to know if it should end its own policy.

## Solution
We configure the CSP header name based on the Rack version in the middleware

It does not concern any part of the OpenAPI spec.

- [x] OAS2
- [x] OAS3
- [x] OAS3.1

### Related Issues
Related to #744

### Checklist
- [ ] Added tests
- [x] Changelog updated
- [x] ~~Added documentation to README.md~~ Does not need documentation

- Set up a Rails application using Rack 3.
- Set a CSP policy in Rails initializer
- Set up rswag in the application
- Try to display the webpage on rswag endpoint, page should be blank with CSP errors in the browser console.


SebouChu:
Don't know where to test this tho, if anyone has an idea :)

woodhull:
We also experienced this!

tagliala:
Hello,

the following change in Rack will not play well with this fix: 

https://github.com/rack/rack/pull/2199/files

SebouChu:
@tagliala I don't see how it can be a problem as Rack::RELEASE is kept the same as Rack::VERSION

tagliala:
> @tagliala I don't see how it can be a problem as Rack::RELEASE is kept the same as Rack::VERSION

I did read the comments at rack/rack#2199, and apparently `Rack::RELEASE` will not be deprecated at all (or it is not planned to be deprecated), so yes, it is not a concern anymore

I was concerned about a future deprecation, and a subsequent removal of that constant

However, `Rack.release` is available and did not change

SebouChu:
@tagliala I'm going to use this to make sure it does not need a change for Rack 4

rswag-ui/lib/rswag/ui/middleware.rb
```diff
@@ -14,7 +19,7 @@ def call(env)
         end
 
         if index_path?(env)
-          return [ 200, { 'Content-Type' => 'text/html', 'Content-Security-Policy' => csp }, [ render_template ] ]
+          return [ 200, { 'Content-Type' => 'text/html', CONTENT_SECURITY_POLICY_HEADER => csp }, [ render_template ] ]
```
Lorel, line 24: As `actionpack` is defined as [dependency](https://github.com/rswag/rswag/blob/master/rswag-ui/rswag-ui.gemspec#L25) of this gem, we can rely on the constants defined there

```suggestion
          return [ 200, { 'Content-Type' => 'text/html', ActionDispatch::Constants::CONTENT_SECURITY_POLICY => csp }, [ render_template ] ]
```
to get rid of the definition of `CONTENT_SECURITY_POLICY_HEADER` above

SebouChu: Thanks @Lorel, made the adjustment!

tagliala: Hello, this is a good suggestion

However, as per gemspec, rswag-ui supports actionpack >= 5.2

https://github.com/rswag/rswag/blob/3013bcd9aaa73d1398bf8b0f4cece17efba91c63/rswag-ui/rswag-ui.gemspec#L25-L26

`ActionDispatch::Constants` has been introduced 2 years ago, probably in 7.1.0

https://github.com/rails/rails/commit/1fd79abb549739c3db06f9fe197ff8836be639c0

I would reconsider this change, and maybe add a note to migrate to ActionDispatch::Constants when all Rails < 7.1 support will be dropped

<details><summary>Copilot analysys</summary>

> This change updates the HTTP header key for Content Security Policy (CSP) from the hardcoded string `'Content-Security-Policy'` to `ActionDispatch::Constants::CONTENT_SECURITY_POLICY`. Specifically:
> 
> **Before:**
> ```ruby
> return [200, { 'Content-Type' => 'text/html', 'Content-Security-Policy' => csp }, [render_template]]
> ```
> **After:**
> ```ruby
> return [200, { 'Content-Type' => 'text/html', ActionDispatch::Constants::CONTENT_SECURITY_POLICY => csp }, [render_template]]
> ```
> 
> ### What does this do?
> - Instead of using the string `'Content-Security-Policy'` as the header key, it uses the constant `ActionDispatch::Constants::CONTENT_SECURITY_POLICY`.
> - This constant is essentially an alias for `'Content-Security-Policy'` in modern Rails versions.
> 
> ---
> 
> ## Is this change good given the supported Rails versions?
> 
> ### 1. **Support for `ActionDispatch::Constants::CONTENT_SECURITY_POLICY`**
> - This constant was [introduced in Rails 7.1](https://github.com/rails/rails/pull/48355).
> - **If rswag-ui supports Rails < 7.1 (e.g., Rails 6.1, 7.0), this change will break compatibility**, as the constant will not exist in those versions.
> - If rswag-ui has dropped support for Rails < 7.1, this change is safe and slightly preferred for consistency and avoiding typos.
> 
> ### 2. **Current rswag Supported Rails Versions**
> - As of early 2024, [rswag](https://github.com/rswag/rswag) supports Rails 6.1, 7.0, and 7.1 ([see gemspec and documentation](https://github.com/rswag/rswag/blob/main/rswag-ui/rswag-ui.gemspec)).
> - **Therefore, this change is NOT safe for all supported rswag users**—it will cause a `NameError` on Rails < 7.1.
> 
> ---
> 
> ## Recommendation
> 
> - **If rswag-ui needs to support Rails < 7.1:**  
>   **Do not make this change** or use a fallback:
>   ```ruby
>   csp_header = defined?(ActionDispatch::Constants::CONTENT_SECURITY_POLICY) ?
>     ActionDispatch::Constants::CONTENT_SECURITY_POLICY : 'Content-Security-Policy'
>   return [200, { 'Content-Type' => 'text/html', csp_header => csp }, [render_template]]
>   ```
> - **If rswag-ui only supports Rails >= 7.1:**  
>   The change is good and more robust.
> 
> ---
> 
> ### **Summary**
> 
> **This change is not safe if rswag-ui supports any Rails version lower than 7.1.**  
> If you want to support older versions, use a conditional fallback. If you only support Rails 7.1 and above, the change is fine.
</details>

Notes on Copilot analysis:
- The link to the PR is completely wrong
- The versions supported by rswag ui are wrong

SebouChu: Oh yeah, now I remember, I think that was why I didn't use ActionPack constant, due to this minimal version. @Lorel Are you okay with me reverting the last change?

franzliedke:
This is the correct fix from my perspective, except for the small difficulty with older Rails versions, where the constant is not present. (Should be fixable by using a custom constant here that falls back the upper-case variant when the Rails constant is not defined.)

In the meantime, here's our generic patch for anybody with similar problems on a Rails app with Rack 3:

```ruby
class RswagCSPDisableMiddleware
  def initialize(app)
    @app = app
  end

  def call(env)
    @req = Rack::Request.new(env)
    _, headers, = response = @app.call(env)

    if headers.key?("Content-Security-Policy")
      headers["content-security-policy"] = headers.delete("Content-Security-Policy")
    end

    response
  end
end

Rails.application.config.middleware.insert_after(
  ActionDispatch::ContentSecurityPolicy::Middleware,
  RswagCSPDisableMiddleware
)
```

franzliedke:
Interestingly, the [Rack 3 upgrade guide](https://github.com/rack/rack/blob/v3.1.16/UPGRADE-GUIDE.md) claims that using all-lower-case header names should give you full compatibility with Rack 2 and 3. Is this not correct in the Rails context?

ShamoX:
Would it be nice to:

1. Add some documentation about the usage of CSP (for the non-initiated, that would activate the config.content_security_policy and face a blank page or non-working API call)?
2. Remove the need for 'unsafe-inline' altogether?

Ok, this last point should not be discussed here. But is there a reason why it has been made? Is it “only” for the simplicity of not having several other script/style files?

rswag-ui/lib/rswag/ui/middleware.rb
```diff
@@ -15,7 +15,7 @@ def call(env)
         end
 
         if index_path?(env)
-          return [200, { 'Content-Type' => 'text/html', 'Content-Security-Policy' => csp },
+          return [200, { 'Content-Type' => 'text/html', ActionDispatch::Constants::CONTENT_SECURITY_POLICY => csp },
```
tagliala: `rswag_ui` supports actionpack and railties >= 5.2

`ActionDispatch::Constants::CONTENT_SECURITY_POLICY` was introduced in Rails 7.1.0.
The ActionDispatch::Constants module was created on Jul 27, 2023 (commit rails/rails@1fd79ab), with the CSP constants added the next day in commit rails/rails@d1381dc.

This should mirror Rails' implementation otherwise it will break on older Rails versions

SebouChu: @tagliala how about now?

tagliala: This seems to be a better choice at the moment

PR #776, opened 2024-08-27
Remove `.ruby-version` lock file

AlexWayfer:
We don't need for it in development environment.

And we should test gems among all supported Ruby versions.

## Problem

RuboCop and other tools requires me to install specific (pretty old) version of Ruby. But I don't have to, for development process.

## Solution

Just remove a specific-version-lock-file. All supported Ruby versions will be tested in CI.

### Checklist
- [ ] Added tests
- [ ] Changelog updated


jtannas (approved):

jtannas:
@romanblanco Apologies but I merged this without thinking about the changelog needing to be updated.
I've emailed you about getting back into sync about the status of the project.

PR #790, opened 2024-10-09
feat: replace OpenStruct by Struct in specs

### LINKED ISSUE #789: [REQUEST] Add ostruct to the gemspec of rswag-ui
ostruct will be removed from the ruby standard library starting from 3.5.0. When booting a Rails app using rswag on Ruby 3.3, we get the following deprecation warning: 
```
warning: /Users/tao/.rbenv/versions/3.3.5/lib/ruby/3.3.0/ostruct.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
Also please contact the author of rswag-ui-2.15.0 to request adding ostruct into its gemspec.
```

Tao-Galasse:
fixes https://github.com/rswag/rswag/issues/789

Fixes a deprecation warning for future Ruby versions

### DIFF
--- a/rswag-specs/spec/rswag/specs/configuration_spec.rb
+++ b/rswag-specs/spec/rswag/specs/configuration_spec.rb
@@ -3,11 +3,17 @@
 require 'rswag/specs/configuration'
 require 'climate_control'
 
+RspecConfig = Struct.new(
+  :openapi_root, :openapi_specs, :openapi_format, :rswag_dry_run,
+  :openapi_strict_schema_validation, :openapi_all_properties_required, :openapi_no_additional_properties,
+  keyword_init: true
+)
+
 RSpec.describe Rswag::Specs::Configuration do
   subject { described_class.new(rspec_config) }
 
   let(:rspec_config) do
-    OpenStruct.new(
+    RspecConfig.new(
       openapi_root: openapi_root, openapi_specs: openapi_specs,
       openapi_format: openapi_format, rswag_dry_run: rswag_dry_run,
       openapi_all_properties_required: openapi_all_properties_required,
--- a/rswag-specs/lib/rswag/specs/response_validator.rb
+++ b/rswag-specs/lib/rswag/specs/response_validator.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'active_support/core_ext/enumerable'
 require 'active_support/core_ext/hash/slice'
 require 'json-schema'
 require 'json'
--- a/rswag-specs/spec/rswag/specs/openapi_formatter_spec.rb
+++ b/rswag-specs/spec/rswag/specs/openapi_formatter_spec.rb
@@ -5,6 +5,9 @@ require 'ostruct'
 
 module Rswag
   module Specs
+    Metadata = Struct.new(:metadata)
+    Group = Struct.new(:group)
+
     RSpec.describe OpenapiFormatter do
       subject { described_class.new(output, config) }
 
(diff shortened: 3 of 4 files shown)


dawidof:
Maybe it's better to change from `OpenStruct` to `Struct` or `Data`?

Tao-Galasse:
> Maybe it's better to change from `OpenStruct` to `Struct` or `Data`?

Probably, yes; but in would require more refactoring. What I did was a quick fix to get rid of the deprecation warning and could be pushed as a minor release. 
A refactoring to replace `OpenStruct` with `Struct` could indeed be worked on and proposed as a mid release?

Spone:
FYI `OpenStruct` is only used [in spec files](https://github.com/search?q=repo%3Arswag%2Frswag%20openstruct&type=code) currently, so I think it will be pretty straightforward to change to `Struct`. I acknowledge it can be done separately, though.

rswag-specs/spec/rswag/specs/configuration_spec.rb
```diff
@@ -3,11 +3,16 @@
 require 'rswag/specs/configuration'
 require 'climate_control'
 
+RspecConfig = Struct.new(
+  :openapi_root, :openapi_specs, :openapi_format, :rswag_dry_run,
+  :openapi_strict_schema_validation, :openapi_all_properties_required, :openapi_no_additional_properties
+)
```
Spone: You need to pass `keyword_init: true` if you want to use keyword arguments when you initialize with `RspecConfig.new` (see [Keyword Argument](https://ruby-doc.org/core-3.1.0/Struct.html) section in Ruby docs)

```suggestion
RspecConfig = Struct.new(
  :openapi_root, :openapi_specs, :openapi_format, :rswag_dry_run,
  :openapi_strict_schema_validation, :openapi_all_properties_required, :openapi_no_additional_properties,
  keyword_init: true
)
```

Tao-Galasse:
@Spone You were right, I didn't realize it was only used in spec files.

I updated the PR; the dependency to OpenStruct is no more added to the gemspec, and the occurrences to `OpenStruct` have been replaced in favor of `Struct`.

I had to `require 'active_support/core_ext/enumerable'` in `rswag-specs/lib/rswag/specs/response_validator.rb`, because the `exclude?` method was not defined on Hash and the tests crashed.

rswag-specs/spec/rswag/specs/configuration_spec.rb
Tao-Galasse: Looks like you're right; curiously, the tests passed without it though 🤔 Could it be thanks to ruby 3+ infering keyword arguments?

Spone: Yes, that's weird.

dawidof:
@Spone could you make CR again, please?

Spone (approved):
Looking good!

BookOfGreg (approved):

dawidof:
@Tao-Galasse could you resolve conflicts?

Tao-Galasse:
@dawidof Done! :)

Tao-Galasse:
Hi @BookOfGreg @romanblanco @jtannas, any thoughts on this? :) 🙏

francois-ferrandis:
Dear maintainers,

I hate to be that person, but I would love some update on this. This issue is the only reason we keep `ostruct` as a dependency in our project. The fix here seems quite straightforward.

I completely understand that project stewardship is time consuming and respect your time. Just wanted to give you a little nudge, just in case.

Have a great day. :pray:

romanblanco (approved):

harsh183:
will this be included in version 3 official release? I don't see this change in [2.17.0](https://github.com/rswag/rswag/releases/tag/2.17.0) or [3.0.0.pre](https://github.com/rswag/rswag/releases/tag/3.0.0.pre)

romanblanco:
@harsh183, I plan to include it in the final v3 release

harsh183:
@romanblanco that sounds fantastic!

harsh183:
@romanblanco, do you have a guess on when the next/final v3 release might be? I can remove the `ostruct` requirement from my `Gemfile` after that?

PR #809, opened 2025-01-17
Fix filepath in README

koddsson:
## Problem
The README references a file path that doesn't exist

## Solution
Changing the file path in the README.

_I removed the rest of the PR template since I felt like it wasn't applicable to a docs change in the README_


jtannas (approved):

jtannas:
@romanblanco Apologies but I merged this without thinking about the changelog needing to be updated.
I've emailed you about getting back into sync about the status of the project.

PR #816, opened 2025-01-31
Chore: ruby 3.3.7, dockerfile to Node 22, Rails 7.2.0, Hadolint

BookOfGreg:
## Problem
Bump to latest versions

## Solution
Use nodesource for newer node.
Same as #812 but for V3 and not rails 8 yet


test-app/db/schema.rb
```diff
@@ -16,7 +16,7 @@
     t.text "content"
     t.string "thumbnail"
     t.string "status"
-    t.datetime "created_at", precision: nil
-    t.datetime "updated_at", precision: nil
+    t.datetime "created_at"
+    t.datetime "updated_at"
```
BookOfGreg, line 20: From running the tests.

jtannas:
@BookOfGreg Apologies but I approved and merged a PR that removed the `.ruby-version` file.
https://github.com/rswag/rswag/pull/776

What're your thoughts on the reasoning there?

P.S. I've really been liking [`asdf`](https://asdf-vm.com/) for managing multiple runtime versions in a single project (eg. [ruby](https://github.com/asdf-vm/asdf-ruby)+[nodejs](https://github.com/asdf-vm/asdf-nodejs)+[sqlite](https://github.com/cLupus/asdf-sqlite)+...)

BookOfGreg:
That's fine, can absolutely remove the ruby-version.

PR #815, opened 2025-01-31
V3 rubocops

BookOfGreg:
## Problem
Like #813 but on v3 branch, and more of them.

## Solution
Cops split more or less by commit. Do feel free to cherry-pick or reject or rebase to your preference.
Last time I had focused on the rswag-specs folder as that's where most of the action is. This time I apply to the wider repo.

PR #823, opened 2025-02-07
v3/rubocop/rubocop checks on PRs

jtannas:
Extends #815 with
- Automatic checking of rubocop via Github Actions
- Adding relevant rubocop extensions (eg. rails, rspec, etc...)
- Autocorrecting offenses where safe autocorrection exists
- Using a `rubocop-todo.yml` for offenses that must be manually corrected or with unsafe autocorrection


.rubocop.yml
```diff
@@ -0,0 +1,23 @@
...
+# RuboCop will start looking for the configuration file in the directory
+# where the inspected file is and continue its way up to the root directory.
+#
+# See https://docs.rubocop.org/rubocop/configuration
+require:
+  - rubocop-capybara
+  - rubocop-rails
+  - rubocop-rake
+  - rubocop-rspec
+  - rubocop-rspec_rails
```
BookOfGreg, line 19: We could also add rubocop-performance

jtannas: Done :)

BookOfGreg (approved):
Great extension to the original PR. I wasn't sure if the rubocops would be a help or a hinder, seems like we're choosing to adopt it as a group. Nice.

jtannas:
> Great extension to the original PR. I wasn't sure if the rubocops would be a help or a hinder, seems like we're choosing to adopt it as a group. Nice.

Thanks :bow: 
I'll merge it into your PR now so that it's a single merge into `v3`

Issue #297, opened 2020-04-10
Feature request: Auto generate docs


jamesla:
I think a great feature would be if we could make it so the swagerised docs were generated each time we ran the rspec test files changed. 

This way they would always be up to date and would reduce the chance of a deploy going out where somebody has forgotten to generate the latest docs.

BookOfGreg:
rswag-specs is implemented as an rspec formatter, if you want it to run everytime you can add the `--format Rswag::Specs::SwaggerFormatter` flag to the rspec command.
I believe you can set that flag multiple times if you have other formatters you also want.

Does this help? Or do you have an idea that would be more useful or easy to discover?

jamesla:
Yep that works perfectly thank you.  (here's a great place for it, like most people the first place I search for things like this is in the issues of the github).

PR #844, opened 2025-06-24
Add Rspec formatter info to README

Floppy:
After I found this info in #297, I thought it was worth including in the README!


Floppy:
still relevant... maybe.

PR #852, opened 2025-08-26
Fix rack status which needs to be an integer

dmathieu:
The Rack status needs to be an integer, and we return a string.
See https://www.rubydoc.info/github/rack/rack/file/SPEC.rdoc#label-The+Status

This is causing issues in rack methods such as [server_error](https://github.com/rack/rack/blob/main/lib/rack/response.rb#L187), as strings can't be compared with integers.

### DIFF
--- a/rswag-api/lib/rswag/api/middleware.rb
+++ b/rswag-api/lib/rswag/api/middleware.rb
@@ -28,7 +28,7 @@ module Rswag
           body = unload_openapi(filename, openapi)
 
           return [
-            '200',
+            200,
             headers,
             [body]
           ]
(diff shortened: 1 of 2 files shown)


MrLuna12:
I am currently dealing with the same issue. Do we have a timeline on when this pr will be reviewed and merged?
cc: @BookOfGreg @romanblanco

dmathieu:
I see some new commits landed on the main branch. Ping, as this change is trivial, and would be really welcome.

romanblanco:
Hi @dmathieu, thanks for the ping!

The change seems fine to me, I only don't see that tests would run for the PR as they should.
Would you mind rebasing the branch?

romanblanco (approved):

romanblanco:
No need, tried locally and tests are passing.

PR #876, opened 2026-01-15
Remove upper bound dependency on Rails gems

byroot:
I'd like to test our app against Rails edge, but this gem is preventing it because it has a pessimistic constraint on Rails gems.

There is no reason to believe 8.2 will break this gem, but even if it does, by allowing edge testing you are likely to receive a fix long in advance of the final release, which is preferable.

### DIFF
--- a/rswag-api/rswag-api.gemspec
+++ b/rswag-api/rswag-api.gemspec
@@ -19,8 +19,8 @@ Gem::Specification.new do |s|
 
   s.files = Dir['{lib}/**/*'] + %w[MIT-LICENSE Rakefile]
 
-  s.add_dependency 'activesupport', '>= 5.2', '< 8.2'
-  s.add_dependency 'railties', '>= 5.2', '< 8.2'
+  s.add_dependency 'activesupport', '>= 5.2'
+  s.add_dependency 'railties', '>= 5.2'
 
   s.add_development_dependency 'simplecov', '=0.21.2'
 end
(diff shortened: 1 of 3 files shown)


BookOfGreg:
Hi @byroot , I recognise your from your contributions, thanks for coming by! I'll mention we're a small group here and we keep Rails pinned because occasionally we have to handle day1 breakages (hello connection_pool) and it's best for us to be conservative in general.
Do you know of a good example of a gem that manages to balance conservative gem pinning with experimentation/edge testing? I would like to follow a good example if you knew of one.

byroot:
> Do you know of a good example of a gem that manages to balance conservative gem pinning with experimentation/edge testing?

There is no middle ground. Either upper versions are allowed, or either they aren't. Right now I have no choice but to run a fork of your project.

> I'll mention we're a small group here and we keep Rails pinned because occasionally we have to handle day1 breakages (hello connection_pool) and it's best for us to be conservative in general.

Pinning is more work for you. If you don't pin there is 99% chance you won't have anything to do when Rails release a new version, and 1% change you'll get day 1 pull requests from people like me who run Rails edge.

If you keep it pinned, you'll get day 1 complaints from people like me who run Rails edge or are trying to test release candidates.

Which is more work do you think?

But also, the situation you mention is no big deal, every one of your users have a lockfile, gems don't upgrade themselves and people can keep the dependency locked temporarily. But more importantly, it's specifically Rails gem I'm talking about. No-one inadvertently upgrade Rails.

PR #880, opened 2026-04-22
add `writeOnly` support for response validation

dgmora:
I've used RSwag for a while, but first time contributor.

Disclaimer: Implemented with AI, but I did multiple tweaks. It makes sense to me, but I don't know the internals that well.

Happy to do changes or receive suggestions. 

As I understand this, json schema validation cannot cover `writeOnly`, and it is an extra layer we have to add. To do the minimum amount of changes, what we do is just remove the properties that have `writeOnly` and are part of the `required` list.

### DIFF
--- a/rswag-specs/lib/rswag/specs/extended_schema.rb
+++ b/rswag-specs/lib/rswag/specs/extended_schema.rb
@@ -4,11 +4,64 @@ require 'json-schema'
 
 module Rswag
   module Specs
+    class ResponseRequiredAttribute < JSON::Schema::RequiredAttribute
+      def self.validate(current_schema, data, fragments, processor, validator, options = {})
+        return super unless data.is_a?(Hash)
+
+        filtered_schema = remove_write_only_required_properties(current_schema, validator)
+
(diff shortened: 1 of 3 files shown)


Drowze:
I didn't know about `readOnly` / `writeOnly` before this PR. Looks like a very useful feature - looking to use it with rswag soon 😄 

For instance we have 100+ schemas defined under components/schemas and a lot of similarity between request/response schemas, this would allow us to refactor quite a bit!

Drowze:
I did some research on this and thought it was worth mentioning: In OpenAPI Specification v3.0, the `readOnly` and `writeOnly` properties are explicitly specified as being used for validation. However, starting with v3.1, the spec shifted to fully adopt JSON Schema and, since `readOnly`/`writeOnly` don’t have significance in pure JSON Schema, the OpenAPI v3.1 spec now describes them as *annotations* that <ins>MAY</ins> be used for validation.

That being said, it does seem that a few other OpenAPI-related tools also use these fields for validation. I like the idea and would like this on rswag too. :)

